PR #3, opened 2026-04-06
feat: add to cache callable object for eager op

rebel-chanheo:
## Summary of Changes

Cache the compiled callable returned by `torch.compile(model, backend="rbln", ...)` for eager ops, keyed by `(model identity, dynamic, device, options)`, so repeated eager-op calls reuse the wrapper instead of re-entering the full `torch.compile` path every time.

- New `torch_rbln/_internal/compile_cache.py`
  - `compile_rbln_cached(model, *, dynamic, options, device_cache_key)` — cache hit returns previously compiled callable; miss calls `torch.compile(..., backend="rbln", ...)` and stores the result.
  - `_IdentityKey` wraps the model with `id()`-based hash and `is`-based equality, and holds a strong reference to prevent id reuse after GC. `_freeze_cache_value` normalizes `Mapping`/`Sequence`/`set` into deterministic tuples.
  - `clear_rbln_compile_cache()` exposed; thread-safe via double-checked locking.
- `torch_rbln/_internal/monkey_patches.py`
  - Also patches `torch._dynamo.reset` so a dynamo reset flushes the RBLN cache (no stale wrappers).
  - Stashes originals so `remove_all_patches()` actually restores `torch.compile` and `torch._dynamo.reset`.
- Call-site migrations (raw `torch.compile(..., backend="rbln")` → `compile_rbln_cached(..., device_cache_key=...)`)
  - `kernels/custom_transpose.py`, `kernels/sdpa.py` (2 fwd + 4 bwd)
  - `register_custom_ops.py` — softmax, pow, paged_attn_{prefill,decode}, paged_causal_attn_{prefill,decode}. The four paged-attn paths previously instantiated a fresh `nn.Module` per call; they are now promoted to module-level singletons (`_paged_*_op_module = ...().eval()`) so the identity cache key actually hits.
  - Codegen template (`tools/codegen/generators/general_template.py`) and header (`tools/codegen/codegen.py`) routed through `compile_rbln_cached`.
- Tests in `test/rbln/test_torch_compile_patch.py` cover: steady-state cache reuse, distinct models with colliding `id()`, singleton reuse across paged-attn paths, recompile after `torch._dynamo.reset()`, and `remove_all_patches()` actually restoring originals.

---

## Related Issues / Tickets

* Related to #

---

## Type of Change

- [x] `core` — Changes to `torch.compile` monkey patch and op registration

---

## Affected Modules

- [x] Backend
- [x] Ops/Kernels
- [x] `torch.compile`
- [x] Build/Tools

---

## How to Test

1. `pytest test/rbln/test_torch_compile_patch.py -v` — all new tests should pass.
2. Run a representative inference script and verify a given eager op triggers `torch.compile` only once; subsequent calls hit the cache.
3. Call `torch._dynamo.reset()` during a run and confirm the next eager-op call re-enters `torch.compile`.

---

## Notes

- The cache is process-global and `_IdentityKey` keeps a strong reference to the model. All current call sites feed in module-level singletons (or the codegen template's top-level `op_module_var`), so no leak. **New call sites must pass a singleton module — never a freshly instantiated `nn.Module` per call**, or every call misses the cache and leaks.
- Reviewer checklist
  - [ ] No remaining raw `torch.compile(..., backend="rbln")` outside the cache.
  - [ ] Any newly added op follows the singleton pattern.
  - [ ] Codegen outputs regenerated so generated files import `compile_rbln_cached`.


rebel-chanheo:
## Measured impact of this PR (per-call dispatch overhead)

I instrumented the eager-dispatch path on \`torch.add(a_dev, b_dev)\` and decomposed the per-call cost at multiple layers, before and after applying this PR. Methodology and breakdown below; one number to lead with: **~280 µs eliminated per call**, dominantly because \`torch.compile(...)\` is no longer reinvoked per dispatch.

### Setup

- Op: \`add\` (binary elementwise) via \`torch.add(a, b)\` with both inputs on \`rbln\` device, fp16
- 100 warmup + 1000 timed iterations after first compile (steady state, cache hit only)
- TP=1, \`TORCH_RBLN_DISABLE_FALLBACK=compile_error\`, \`TORCH_RBLN_DEPLOY=ON\`
- Shapes: 1, 10, 100, 1024, 10000, 100000, 1M (median wall-clock, perf_counter)

### Layers measured

- **L0**: \`torch.add(a_dev, b_dev)\` — full path through register_ops wrapper
- **L1**: \`compiled_fn(a_dev, b_dev)\` — pre-compiled wrapper called directly (skips register_ops Python wrapper)
- **L2**: dynamo-inner — bypasses torch.compile/dynamo wrapper, lands on rebel runtime
- Sub-decomposition of L2 via instrumented \`DynamoRuntime.run()\`

### Total per-call (median µs)

| numel | L0 **before** | L0 **after** | Δ | reduction |
|------:|-----:|-----:|-----:|------:|
| 1024 | 507 | **225** | -282 | -56% |
| 10000 | 504 | **224** | -280 | -56% |
| 100000 | 508 | **230** | -278 | -55% |
| 1M | 521 | **269** | -252 | -48% |

\`L0\` is essentially flat with shape until ~1M, because compute is dominated by Python+dispatch overhead.

### register_ops Python wrapper layer (= L0 - L1)

| numel | wrapper **before** | wrapper **after** | reduction |
|------:|-----:|-----:|------:|
| 1024 | 373 | **92** | -75% |
| 10000 | 368 | **91** | -75% |
| 100000 | 372 | **95** | -74% |
| 1M | 360 | **102** | -72% |

The ~280 µs cut maps directly to skipping \`torch.compile(_op_module, backend=\"rbln\", ...)\` reinvocation on every eager call. The remaining ~92 µs is the rest of the wrapper body (\`is_cpu_fallback_cases\`, \`extract_tensors\`, \`broadcast_args_general\`, \`prepare_args_for_contiguous\`, etc.).

### Full breakdown after applying this PR (1024 fp16 add)

| Component | µs | share |
|---|---:|---:|
| register_ops Python wrapper (cache hit) | 92 | 41% |
| torch.compile / dynamo wrapper | 7 | 3% |
| DynamoRuntime jacket (eager_helper, env, capture) | 37 | 16% |
| sub-runtime (Python prepare_inputs + ptr collect + 2x C++ prepare + alloc) | 22 | 10% |
| \`_runtime_handle.run()\` (C++ device call) | 67 | 30% |
| **L0 total** | **225** | 100% |

The C++ \`_runtime_handle.run()\` (~67 µs) is the hard floor for any single eager op — no Python-side change can reduce it.

rebel-seungchul (approved):
LGTM

rebel-clee (approved):
LGTM

PR #6, opened 2026-04-08
refactor: update zero op to reduce overhead

rebel-chanheo:
## Summary of Changes

Optimize `aten::zero_` for RBLN tensors by replacing the CPU fallback path with lazy zero marking via `VMemorySyncState::EMPTY_INIT_WITH_ZERO`.

**Before:** `zero_()` allocated host memory, filled it with zeros (memset), then transferred the entire buffer to the device (H2D) — every time, unconditionally.

**After:** `zero_()` sets a VMemory flag in O(1). Actual zero data is materialized lazily only when needed, and the cost depends on the access pattern:

### Performance benefit by access pattern

| Scenario | Before (CPU fallback) | After (lazy mark) | Savings |
|---|---|---|---|
| **Write-first (KV-cache)** — NPU overwrites the tensor before reading | host malloc + memset + H2D (all wasted) | flag only, zero transfer **skipped entirely** | **host alloc + memset + H2D eliminated** |
| **Host read** (`.cpu()`, `.item()`) | host malloc + memset + H2D + D2H | host malloc + memset (direct, no device round-trip) | H2D + D2H eliminated |
| **NPU read** (tensor used as operand) | host malloc (permanent) + memset + H2D | temp buffer + memset + H2D + **immediate free** | Persistent host allocation eliminated |
| **Re-zeroing** (repeated `zero_()`) | full malloc + memset + H2D each time | flag set only; previous host memory **freed** | All repeated costs eliminated |

### Primary target: KV-cache initialization

KV-cache tensors are created with `torch.zeros()`, then fully overwritten during prefill. The old path allocated GBs of host memory and transferred zeros to the device — all of which was immediately discarded when prefill wrote real data. The new path eliminates this entirely: zero marking is O(1), and the subsequent NPU write transitions directly to `PHYSICAL_VIEW_IS_LATEST` without any zero transfer.

### Implementation

- `c10::rbln::mark_zeros()` — C++ function that calls `rbln_mark_zeros(vaddr)` to set VMemory state
- `torch_rbln._C._mark_zeros` — Python binding exposed via pybind11
- `custom_zero__rbln()` — Python eager op now calls `_mark_zeros(self.data_ptr())` instead of CPU fallback
- Early return for empty tensors (`numel() == 0`)

### Tests (41 new)

| Category | Count | Coverage |
|---|---|---|
| Basic correctness | 8 | readback (1D/2D/3D), scalar, empty, idempotent, overwrite, return self |
| Arithmetic | 5 | add, mul, in-place add, matmul, comparison |
| Partial write | 5 | slice front/middle, 2D row/column, advanced indexing |
| Views | 7 | reshape, view, transpose, contiguous, slice view, write-through-view, expand |
| Copy | 4 | copy from/into zero, clone isolation, CPU round-trip |
| KV-cache pattern | 3 | full overwrite, partial fill, incremental decode |
| Large tensor | 2 | 1024×1024 readback, large compute |
| Re-zero transitions | 4 | after compute, host read, device write, alternating fill/zero |
| torch.zeros factory | 3 | zeros, zeros_like, zeros + operation |

All 41 tests pass on RBLN hardware.

---

## Related Issues / Tickets

* Related to KV-cache initialization performance

---

## Type of Change

- [ ] `fix` — Bug fix

---

## Affected Modules

- [ ] Backend
- [ ] Device
- [x] Ops/Kernels
- [ ] `torch.compile`
- [ ] Memory
- [ ] Build/Tools
- [ ] Docs

---

1. Run `python -m pytest test/rbln/test_zero_opt.py -v` — all 41 tests should pass
2. Verify KV-cache pattern: `zero_()` → NPU write should not trigger any H2D zero transfer (check runtime logs for absence of zero-buffer transfer)
3. Verify host read: `t.zero_(); t.cpu()` should return all zeros without device round-trip

---

## Checklist

---


rebel-seungchul (approved):
LGTM

PR #13, opened 2026-04-16
feat(copy): wire non_blocking copies through async memcpy bindings

rebel-jonghewk:
## Description

Wire `non_blocking=True` host↔device copies through the async memcpy / synchronize APIs from rebel_compiler #10435.

**1. Async copy bindings**
- **API:** `c10::rbln::memcpy_{h2v,v2h,v2v}_async` + `synchronize(device)` (exposed as `torch.rbln.synchronize()`). Returns a hardware seq handle; sync fallback is transparent.
- **V2V:** same-device → `rbln_memcpy_v2v_async`; cross-device → sync `memcpy_v2v` (host-bounce) fallback.

**2. `_copy_from_rbln` wiring**
- **Blocking:** drain the stream, then run the existing sync copy.
- **Non-blocking:** direct copies take the async fast-path; non-direct copies drain then fall back to sync (prevents back-to-back `non_blocking` copies from racing). Zero-byte short-circuit.

**3. Performance**
- **Result:** non_blocking copy is **1.35x~2.5x** faster than sync (larger for many-small copies).
- **Precondition:** requires a physical view (or `TORCH_RBLN_EAGER_MALLOC=1`). 

| dir | N × bytes | speedup |
|---|---|---|
| D2H | 256 × 4K | 2.49x |
| D2H | 4 × 64M | 1.35x |
| H2D | 64 × 4K | 2.08x |

**4. Tests**
- C++ `RBLNFunctionsTest`: `AsyncMemcpyH2VAndV2H`, `AsyncMemcpyV2V`.
- Python: `test_tensor_copy.py` (`non_blocking` axis), `test_rbln_apis.py::test_synchronize`.

Depends on rebel_compiler #10435.

---

## Type of Change
- [x] `feature`

## Affected Modules
- [x] Memory


rebel-jonghewk:
@rebel-chanheo @rebel-seungchul @rebel-clee can you review this plz?

rebel-clee:
@rebel-jonghewk I pushed a commit to sync the vendored runtime header with `constraints-build-dev.txt`. Two changes:

- `constraints-build-dev.txt`: bumped `rebel-compiler` from `0.10.4.dev41+g8746e1bf.prod` to `0.10.4.dev198+gdf5ca3f6`.
- `third_party/rebel_compiler/include/rebel/runtime/api/rbln_runtime_api.h`: added the async declarations (`rbln_memcpy_{v2h,h2v,v2v,v2v_multi}_async`, `rbln_transfer_wait`, `rbln_device_synchronize`) to match.

Happy to repin if you had a different version in mind.

rebel-jonghewk:
> @rebel-jonghewk I pushed a commit to sync the vendored runtime header with `constraints-build-dev.txt`. Two changes:
> 
> * `constraints-build-dev.txt`: bumped `rebel-compiler` from `0.10.4.dev41+g8746e1bf.prod` to `0.10.4.dev198+gdf5ca3f6`.
> * `third_party/rebel_compiler/include/rebel/runtime/api/rbln_runtime_api.h`: added the async declarations (`rbln_memcpy_{v2h,h2v,v2v,v2v_multi}_async`, `rbln_transfer_wait`, `rbln_device_synchronize`) to match.
> 
> Happy to repin if you had a different version in mind.

@rebel-clee  Thanks alot !

rebel-clee:
The new Python-facing surface in this PR — the `non_blocking=True` branch in `_copy_from_rbln` and the new `torch.rbln.synchronize()` API — has no tests under `test/rbln/`. Could we add some?

The conventions for new RBLN backend tests are documented in `docs/TEST_GUIDE.md`, including the Minimal Test Template in the Writing Tests section. For this PR, two existing files fit naturally:

- `test/rbln/test_tensor_copy.py` — `TestTensorCopy.test_tensor_copy_kernel` already parametrizes over copy type × contiguity × dtype, so adding a `non_blocking` axis would cover the new asynchronous copy path.
- `test/rbln/test_rbln_apis.py` — `TestDeviceManagementAPIs` already tests `device_count`, `is_available`, and `current_device`, so a `test_synchronize` would fit there naturally.

Additional non-blocking copy tests would also be welcome.

c10/rbln/RBLNFunctions.cpp
```diff
@@ -319,6 +319,65 @@ void memcpy_v2v(void* rbln_dst_data, const void* rbln_src_data, size_t nbytes) {
...
+  const auto dst_host_ptr = reinterpret_cast<uintptr_t>(cpu_dst_data);
+  const auto size = static_cast<uint64_t>(nbytes);
+  uint64_t handle = 0;
+  RBLN_LOG_DEBUG(
+      "Calling rbln_memcpy_v2h_async: src_vaddr={:#x}, dst_host_ptr={:#x}, size={}", src_vaddr, dst_host_ptr, size);
+  RBLN_CHECK(!::rbln::rbln_memcpy_v2h_async(src_vaddr, dst_host_ptr, size, &handle));
+  RBLN_LOG_DEBUG("V2H async dispatched (handle={}, 0=sync fallback)", handle);
+}
+
+void memcpy_v2v_async(void* rbln_dst_data, const void* rbln_src_data, size_t nbytes) {
```
rebel-clee, line 359: `memcpy_v2v` has separate paths for same-device and cross-device copies because `rbln_memcpy_v2v` only handles same-device copies. Does `rbln_memcpy_v2v_async` have the same constraint?

If it does, could `memcpy_v2v_async` similarly split the two cases — `AsyncMemcpyV2V` only covers same-device copies, so an unhandled cross-device copy would currently go undetected.

aten/src/ATen/native/rbln/RBLNCopy.cpp
```diff
@@ -159,23 +159,56 @@ void copy_impl_rbln(const at::Tensor& src, const at::Tensor& dst) {
...
+  const auto dst_device = dst.device();
+  const auto direct_copy = is_direct_copy(src, dst);
+
+  // Async only supported for direct copies (same size, dtype, both contiguous).
+  // Non-direct copies require CPU-side staging which needs synchronous data access.
+  if (!direct_copy) {
+    RBLN_LOG_DEBUG("Non-direct copy, falling back to sync");
+    copy_impl_rbln(src, dst);
+    return;
+  }
```
rebel-clee, line 200: Shouldn't the non-direct fallback in `copy_impl_rbln_async` also drain pending asynchronous transfers before calling `copy_impl_rbln`? The blocking branch in `_copy_from_rbln` does this (`synchronize` then `copy_impl_rbln`), but the fallback skips it. For example:

```py
t1.copy_(t2, non_blocking=True)   # async H2V, still in flight
t1.copy_(t3, non_blocking=True)   # non-direct → sync fallback runs immediately, races the first
```

If so, adding `synchronize` before `copy_impl_rbln` would fix this.

c10/rbln/RBLNFunctions.cpp
rebel-jonghewk: Good catch — `rbln_memcpy_v2v_async` only handles same-device copies, same as the sync version. Mirrored the split in `memcpy_v2v_async`: same-device → `rbln_memcpy_v2v_async`, cross-device falls back to sync `memcpy_v2v` (host-bounce). True async cross-device would need to own the staging buffer past the call's return, which we don't support yet — leaving it for a follow-up.

aten/src/ATen/native/rbln/RBLNCopy.cpp
rebel-jonghewk: You're right, the fallback was racy. Added `c10::rbln::synchronize(rbln_device.index())` before `copy_impl_rbln` in the non-direct branch, matching the pattern in the blocking path of `_copy_from_rbln`.

rebel-jonghewk:
> The new Python-facing surface in this PR — the `non_blocking=True` branch in `_copy_from_rbln` and the new `torch.rbln.synchronize()` API — has no tests under `test/rbln/`. Could we add some?

Added:
- `test/rbln/test_tensor_copy.py::test_tensor_copy_kernel` — new `non_blocking` axis on top of the existing copy-type × contiguity × dtype matrix; sync via `torch.rbln.synchronize()` before the comparison.
- `test/rbln/test_rbln_apis.py::TestDeviceManagementAPIs::test_synchronize` — covers `synchronize()` argument forms (None / int / `torch.device` / str) and post-async-copy correctness.

rebel-clee (approved):
LGTM

test/cpp/core/RBLNFunctionsTest.cpp
```diff
@@ -285,6 +285,72 @@ TEST_F(RBLNFunctionsTest, GetUninitializedMemoryInfo) {
...
+      src_cpu[i] = static_cast<int8_t>((i * 3) % 127);
+    }
+
+    auto src_rbln = c10::rbln::malloc(device_index, nbytes);
+    auto dst_rbln = c10::rbln::malloc(device_index, nbytes);
+    EXPECT_TRUE(src_rbln != nullptr);
+    EXPECT_TRUE(dst_rbln != nullptr);
+
+    c10::rbln::memcpy_h2v_async(src_rbln, src_cpu.data(), nbytes);
+    c10::rbln::memcpy_v2v_async(dst_rbln, src_rbln, nbytes);
```
rebel-chanheo, line 398: This chains memcpy_h2v_async → memcpy_v2v_async → memcpy_v2h_async over overlapping device areas (src_rbln, then dst_rbln) with no synchronize / rbln_transfer_wait between dispatches. Per the ordering contract in rbln_runtime_api.h, issuing a copy that touches a still-pending area before its wait is exactly what's disallowed (finalize reads the area's MRU command-buffer entry). This passes only because nbytes = 4096 is 64-aligned, so no unaligned D2H bounce occurs. So either chaining is actually safe — in which case the header contract is over-stated — or this is masking a latent bug. Could we add an unaligned variant (e.g. nbytes = 4097 or an odd shape) so the test actually exercises the contract?

aten/src/ATen/native/rbln/RBLNCopy.cpp
```diff
@@ -181,23 +181,59 @@ void copy_impl_rbln(const at::Tensor& src, const at::Tensor& dst) {
...
   if (non_blocking) {
-    if (c10::rbln::is_fallback_disabled("non_blocking_copy")) {
-      RBLN_CHECK(
-          false,
-          "Non-blocking copy is not supported on RBLN devices. "
-          "To enable fallback to blocking copy, remove 'non_blocking_copy' from `TORCH_RBLN_DISABLE_FALLBACK`.");
-    } else {
-      RBLN_WARN_ONCE("Non-blocking copy is not supported, falling back to blocking copy");
-    }
+    copy_impl_rbln_async(src, dst);
```
rebel-chanheo, line 238: Ordering-contract hazard: back-to-back direct non_blocking copies on the same device area

The contract in rbln_runtime_api.h states: between an async dispatch and its rbln_transfer_wait, do not issue any other copy that touches the same device area — an unaligned D2H bounce would otherwise finalize against the wrong host address.

We drain before sync copies (the else branch below) and before non-direct non_blocking copies (the !direct_copy fallback in copy_impl_rbln_async), but two direct non_blocking copies touching the same area are dispatched async with no drain in between. The wrappers discard the per-op handle and rely solely on the device-wide synchronize(), so nothing serializes them. Reachable from the public API:

```
a = t.to("cpu", non_blocking=True)   # async V2H on t's area, still pending
b = t.to("cpu", non_blocking=True)   # touches the same src area before a's wait
torch.rbln.synchronize()             # a's finalize may bounce to the wrong host addr (unaligned case)
```
 
The C++ async tests chain H2V→V2V→V2H over overlapping areas but use 4096B (aligned) buffers, so the unaligned-bounce hazard never fires and a regression here wouldn't be caught.

How is the contract enforced for direct↔direct same-area non_blocking copies? Options: track the discarded handle per device-area and drain on conflict, or drain before an async dispatch that targets a still-pending area. An unaligned-size + back-to-back same-area test would also lock this down.

rebel-chanheo (changes requested):

aten/src/ATen/native/rbln/RBLNCopy.cpp
```diff
@@ -181,23 +181,59 @@ void copy_impl_rbln(const at::Tensor& src, const at::Tensor& dst) {
...
+  } else {
+    // Drain pending async transfers before sync copy so the vmem state machine is consistent.
+    const auto rbln_device = src.device().is_privateuseone() ? src.device() : dst.device();
```
rebel-chanheo: `rbln_device` resolves to a single device (`src` when both are privateuseone). For a **cross-device V2V**, this drains only the src device — pending transfers on the **dst** device aren't flushed before the sync copy. Narrow edge case, but the drain should cover the device(s) actually involved in the copy.

aten/src/ATen/native/rbln/RBLNCopy.cpp
```diff
@@ -181,23 +181,59 @@ void copy_impl_rbln(const at::Tensor& src, const at::Tensor& dst) {
...
+    RBLN_LOG_DEBUG("No bytes to copy");
+    return;
+  }
+
+  if (src_device.is_cpu() && dst_device.is_privateuseone()) {
+    RBLN_LOG_DEBUG("Async CPU -> RBLN");
+    c10::rbln::memcpy_h2v_async(dst.data_ptr(), src.data_ptr(), nbytes);
+  } else if (src_device.is_privateuseone() && dst_device.is_cpu()) {
+    RBLN_LOG_DEBUG("Async RBLN -> CPU");
+    c10::rbln::memcpy_v2h_async(dst.data_ptr(), src.data_ptr(), nbytes);
```
rebel-chanheo, line 223: non_blocking V2H is dispatched async regardless of whether the CPU dst is pinned. The footgun is narrow but real: the **returned CPU tensor** is plain host memory, so reading it does **not** dispatch through RBLN and nothing drains the pending transfer (this PR's own tests call `torch.rbln.synchronize()` before reading, confirming there's no auto-drain on host read). CUDA sidesteps this by making non_blocking copies to/from **pageable** host synchronous. Contract-following code is fine, but CUDA-habit code like `x = t.to("cpu", non_blocking=True); use(x)` (no `synchronize`) will race here. Consider gating async on pinned host memory, or documenting the stricter requirement.

c10/rbln/RBLNFunctions.cpp
```diff
@@ -322,6 +322,78 @@ void memcpy_v2v(void* rbln_dst_data, const void* rbln_src_data, size_t nbytes) {
...
+  RBLN_LOG_DEBUG(
+      "Calling rbln_memcpy_v2v_async: src_vaddr={:#x}, dst_vaddr={:#x}, size={}", src_vaddr, dst_vaddr, size);
+  RBLN_CHECK(!::rbln::rbln_memcpy_v2v_async(src_vaddr, dst_vaddr, size, &handle));
+  RBLN_LOG_DEBUG("V2V async dispatched (handle={}, 0=sync fallback)", handle);
+}
+
+void synchronize(c10::DeviceIndex device_index) {
+  RBLN_LOG_DEBUG("Synchronizing device {}", static_cast<int>(device_index));
+  check_device_index(device_index);
+  const auto torch_device_id = static_cast<uint32_t>(static_cast<unsigned char>(device_index));
```
rebel-chanheo: Every other function in this file converts the device index via the shared `to_device_id()` helper (e.g. L212), but `synchronize()` open-codes `static_cast<uint32_t>(static_cast<unsigned char>(device_index))`. They produce the same value today (`to_device_id` is currently identity), so this is **not a live bug** — but if logical→physical remapping is ever added to `to_device_id`, `synchronize()` would silently diverge from every other entry point. Suggest using `to_device_id(device_index)` here for consistency / future-proofing.

aten/src/ATen/native/rbln/RBLNCopy.cpp
rebel-jonghewk: Moved into the runtime (rebellions-sw/rebel_compiler#11343): sync copies drain in-flight async transfers on the device they touch; cross-device v2v = v2h(src)+h2v(dst) drains both. Torch-layer drain removed.

aten/src/ATen/native/rbln/RBLNCopy.cpp
rebel-jonghewk: Fixed in rebellions-sw/rebel_compiler#11343: `WaitAndFinalize` now finds the entry by seq instead of the MRU `back()`, so back-to-back same-area copies each finalize correctly. No torch-layer serialization needed.

aten/src/ATen/native/rbln/RBLNCopy.cpp
rebel-jonghewk: RBLN has no pinned-host allocator, so pinned-gating would kill async V2H entirely. Instead the runtime is now safe for in-flight re-copies (rebellions-sw/rebel_compiler#11343), and the `rbln_runtime_api.h` contract documents that dst is valid only after the wait (standard non_blocking semantics). Documented rather than forced-sync.

test/cpp/core/RBLNFunctionsTest.cpp
rebel-jonghewk: Added unaligned (4097B) variants + a same-area back-to-back unaligned V2H test + a sync-drains-pending-async test (the latter in rebellions-sw/rebel_compiler#11343). The chained case is safe; the real hazard (same-area back-to-back) is covered by the per-handle finalize.

c10/rbln/RBLNFunctions.cpp
rebel-jonghewk: Good catch — switched to `to_device_id(device_index)` to match the other entry points and future-proof against a logical→physical remap.

rebel-chanheo (approved):
LGTM!

constraints-build-dev.txt
```diff
@@ -1,2 +1,2 @@
 # Build/dev dependency constraints (single source of truth).
-rebel-compiler==0.10.5.dev109+ga425966d.prod
+rebel-compiler==0.10.5.dev187+g95f45d1f
```
rebel-jonghewk:
```suggestion
rebel-compiler==0.10.5.dev210+g9337de86
```

rebel-jonghewk:
@rebel-chanheo Can we merge this?

PR #16, opened 2026-04-18
fix(ci): avoid duplicate --pattern in RAWCUDADEVICE linter

rebel-chanheo:
## Summary of Changes

RAWCUDADEVICE linter in `.lintrunner.toml` passed two `--pattern` args. The upstream `grep_linter.py` synced in #7 added a check that rejects duplicate `--pattern` at argparse level, so the linter now crashes on any PR that touches `aten/**`, `c10/**`, or `torch/csrc/**`. Fold the two patterns into one regex so the linter invokes cleanly while still matching both `cudaSetDevice(` and `cudaGetDevice(`.

## Related Issues / Tickets

* Related to #7

## Type of Change

- [x] ``fix`` — Bug fix

## Affected Modules

- [x] Build/Tools

## How to Test

1. On any file matching `c10/**` (e.g. `c10/CMakeLists.txt`):
   ``python3 tools/linter/adapters/grep_linter.py --pattern='cuda(Set|Get)Device\(' --linter-name=RAWCUDADEVICE --error-name='raw CUDA API usage' --error-description='test' -- c10/CMakeLists.txt``
   → exits 0, no argparse error.
2. Against a file containing both calls, the linter reports them as before.
3. CI ``Check code quality`` job passes on this PR.

## Checklist

* [x] PR title follows Conventional Commits format
* [ ] This PR is linked to a corresponding issue
* [x] Linting passes
* [x] All CI tests pass (pending CI run)

## Notes

This only fixes the immediate lint crash. Separately tracking a CI change in ``fsw-integration`` so that the same class of breakage (lint config/adapter drift that diff-only lint doesn't exercise) surfaces in the PR that introduces it.

PR #17, opened 2026-04-18
fix(core): prevent exception escape from noexcept RBLNGuardImpl methods

rebel-chanheo:
## Summary of Changes

`RBLNGuardImpl::uncheckedSetDevice` and `RBLNGuardImpl::deviceCount` are declared `noexcept` (interface contract with `c10::impl::DeviceGuardImplInterface`), but their catch handlers call `RBLN_WARN` which expands into potentially-throwing operations (`format_log_message`, `c10::warn`). A throw from inside a catch handler of a `noexcept` function invokes `std::terminate`.

Add an `RBLN_WARN_NOTHROW` helper in `c10/rbln/RBLNLogging.h` that swallows any throw originating inside the warning path, and switch both call sites to it.

## Related Issues / Tickets

* Related to #16 (surfaced when considering broader-scope lint coverage)

## Type of Change

- [x] ``fix`` — Bug fix

## Affected Modules

- [x] Backend
- [x] Device

## How to Test

1. Build: ``cmake --build build --target c10_rbln`` → succeeds.
2. Targeted lint: ``lintrunner --take CLANGTIDY --paths-cmd='echo c10/rbln/impl/RBLNGuardImpl.cpp'`` → ``ok No lint issues.`` (previously reported ``bugprone-exception-escape`` on ``uncheckedSetDevice``).
3. Runtime: existing guard tests still pass (no behavior change — warnings still logged, any secondary throw from the warning path itself is now contained instead of terminating the process).

## Checklist

* [x] Linting passes
* [x] CI tests pass (pending CI run)

## Notes

``deviceCount`` was not flagged by the current CI configuration because the file has been unchanged since the initial commit and therefore never appears in diff-based lint against ``origin/main``. The same shape is preserved preventively here so both methods are safe under the ``noexcept`` contract.

### DIFF
--- a/.lintrunner.toml
+++ b/.lintrunner.toml
@@ -674,8 +674,7 @@ exclude_patterns = [
 command = [
     'python3',
     'tools/linter/adapters/grep_linter.py',
-    '--pattern=cudaSetDevice(',
-    '--pattern=cudaGetDevice(',
+    '--pattern=cuda(Set|Get)Device\(',
     '--linter-name=RAWCUDADEVICE',
     '--error-name=raw CUDA API usage',
     """--error-description=\
--- a/c10/rbln/RBLNLogging.h
+++ b/c10/rbln/RBLNLogging.h
@@ -231,6 +231,22 @@ C10_RBLN_API int get_scope_depth();
   } while (0)
 
 /**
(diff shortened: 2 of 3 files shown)


rebel-clee (approved):
LGTM

PR #18, opened 2026-04-20
fix(build): support scikit-build-core packaging in rebel_compiler

rebel-chanheo:
## Summary

- `apply-custom-rebel.sh` assumed rebel_compiler ships `python/gen_requirements.py` and `python/requirements/core.txt`. After rebel_compiler's switch to scikit-build-core, those paths no longer exist, so the script aborted with an error on modern rebel_compiler checkouts.
- Re-running the script when `$REBEL_HOME/.venv` already existed failed because `uv venv` refuses to overwrite an existing venv. Added `--clear` so reruns replace stale environments cleanly.

## Changes

- Make the legacy `gen_requirements.py` + `requirements/core.txt` steps conditional on their presence (pre-scikit-build-core layouts still work).
- Pass `--clear` to `uv venv` so `.venv` gets recreated on rerun.

## Test plan

- [ ] `REBEL_HOME=/path/to/rebel_compiler TORCH_RBLN_BUILD_TYPE=Debug REBEL_DEPLOY=OFF bash tools/apply-custom-rebel.sh --build-only` succeeds on a scikit-build-core rebel_compiler tree.
- [ ] Rerunning the same command on a tree that already has `.venv` still succeeds.
- [ ] Pre-scikit-build-core rebel_compiler tree (if still in use) continues to build.

## Note

Cherry-picked from the previous repository (commit `e42f8a1` on `rebellions-sw/torch-rbln`) during the repo migration to `RBLN-SW/torch-rbln`.

### DIFF
--- a/tools/apply-custom-rebel.sh
+++ b/tools/apply-custom-rebel.sh
@@ -91,8 +91,8 @@ main() {
   # Update the submodules
   git submodule update --init ./ || return $?
 
-  # Create a virtual environment using uv
-  uv venv .venv && . .venv/bin/activate || return $?
+  # Create a virtual environment using uv (--clear replaces any stale .venv from a prior run)
+  uv venv --clear .venv && . .venv/bin/activate || return $?
 
   # Tooling needed before rebel_install.sh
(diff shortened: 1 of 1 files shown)


rebel-clee (approved):
LGTM

PR #26, opened 2026-04-22
fix(build): align external-rebel dev-setup with scikit-build-core layout

rebel-scottlee:
## Summary of Changes

#18 migrated rebel-compiler's Python package to **scikit-build-core**, but `tools/build-with-external-rebel.sh` (invoked by `./tools/dev-setup.sh external --clean`) still assumed the pre-migration layout. On a Python 3.12 venv with a scikit-build-core-built `REBEL_HOME`, the script fails at the version check with:

```
[ERROR] rebel-compiler was not built for Python 312!
```

even though the extension is built and importable. Same root cause — old layout assumption — was duplicated in five more places in the script (PYTHONPATH shim in `install_dependencies` + in `activate_rebel` + in `build_torch_rbln`, the hand-written `rebel_compiler.pth`, and the `gen_requirements.py`-only install path), so fixing only the version check would not have produced a working `import rebel._C`.

### Root cause

Under scikit-build-core:

- The compiled extension `_C.cpython-*.so` lives in **`$REBEL_HOME/python/build/skbuild/`**, not in `python/rebel/`.
- The source tree `python/rebel/` holds only `.py` files and `_C/*.pyi` stubs.
- Imports are wired up by `_rebel_compiler_editable.pth` + `_rebel_compiler_editable.py` (scikit-build-core redirect hook) written to the consumer venv by `pip install -e`.
- Runtime deps come from `pyproject.toml` `[project].dependencies`; `python/gen_requirements.py` and `python/requirements/core.txt` no longer exist.
- Build deps come from `[build-system].requires`: `scikit-build-core>=0.10`, `pybind11>=2.10,<=2.10.4`, `cython`, `setuptools_scm>=8`, `cmake>=3.26`, `ninja`.

The previous script was built for the opposite of all three.

### Fix

`tools/build-with-external-rebel.sh`:

| Function | Change |
|----------|--------|
| `check_rebel_python_version` | Probe `$REBEL_HOME/python/build/skbuild/` first. Keep `python/rebel/` and `rebel/python/rebel/` as diagnostic fallbacks. Same expansion in the "available versions" listing. |
| `install_dependencies` | Drop `export PYTHONPATH="$REBEL_HOME/python:…"` and the hand-written `rebel_compiler.pth`. Prepending the source tree to `PYTHONPATH` shadows the editable install and breaks `import rebel._C`. Clean up any stale `rebel_compiler.pth` from prior runs. |
| `create_activate_rebel_script` | Drop the same `PYTHONPATH` export from generated `activate_rebel` so the shadow doesn't come back on every venv activation. |
| `install_rebel_python_deps` | **Target scikit-build-core only.** Detect the backend via `grep 'scikit_build_core.build'` on `REBEL_HOME/python/pyproject.toml` and fail loudly with an upgrade hint if a pre-migration tree is supplied. Pre-install rebel-compiler's `[build-system].requires` into the active venv (so they survive `uv sync --no-install-project`), then run `uv pip install --no-build-isolation -e "$REBEL_HOME/python"` — deterministic, no PEP 517 isolated-env PyPI fetches per invocation. |
| `build_torch_rbln` | Drop the duplicate `PYTHONPATH` shim; `cmake/FindRebel.cmake` already reads `REBEL_HOME` directly for headers/libs. |
| `verify_installation` | Replace the stale "PYTHONPATH not set correctly" debug hint with one that matches the new flow (missing `_rebel_compiler_editable.pth` / re-run the editable install). |
| `print_summary` | Update the trailing "How to use" block to match the new env (no `PYTHONPATH`). |

---

## Related Issues / Tickets

* Related to #18 (scikit-build-core packaging migration in `rebel_compiler`) — this PR is the torch-rbln-side follow-up for the external-rebel dev-setup.

---

## Type of Change

- [x] `fix` — Bug fix

---

## Affected Modules

- [x] Build/Tools (`tools/build-with-external-rebel.sh`)

---

## How to Test

1. **Prep a Python 3.12 venv with a scikit-build-core-built rebel-compiler** (as produced by current `rebel_install.sh`):
   ```
   export REBEL_HOME=/path/to/external_rebel_compiler
   ls "$REBEL_HOME/python/build/skbuild/_C.cpython-312-*.so"   # present
   ls "$REBEL_HOME/python/rebel/_C.cpython-*.so" 2>/dev/null   # absent (new layout)
   grep build-backend "$REBEL_HOME/python/pyproject.toml"      # scikit_build_core.build
   ```

2. **Before this PR** — the setup fails at the version check:
   ```
   ./tools/dev-setup.sh external --clean
   # [ERROR] rebel-compiler was not built for Python 312!
   ```

3. **After this PR** — the setup completes and the import test passes:
   ```
   ./tools/dev-setup.sh external --clean
   # [INFO] Found matching rebel._C: _C.cpython-312-x86_64-linux-gnu.so
   # [INFO] Installing rebel-compiler build dependencies (scikit-build-core, pybind11, cython)...
   # [INFO] Installing rebel-compiler (editable, --no-build-isolation) from $REBEL_HOME/python...
   #   torch: 2.10.0+cpu
   #   rebel: 0.10.3.dev321+g8763cc92
   #   torch_rbln: 0.1.10.dev12+g…
   # [INFO] Import test PASSED!
   ```

4. **Legacy-layout regression**: if someone points `REBEL_HOME` at a pre-migration rebel-compiler checkout, the script now fails fast with a clear upgrade message rather than silently attempting a fresh Cython compile.

---

## Checklist

* [ ] This PR is linked to a corresponding issue — linked to #18 instead
* [x] Linting passes (shell script only; no project lint impact)
* [ ] All CI tests pass — CI does not currently exercise `dev-setup.sh external`; verified manually
* [x] Relevant documentation has been updated — `print_summary` and error-hint text brought in line with the new flow

---

## Notes

- Only `tools/build-with-external-rebel.sh` is modified. `pyproject.toml` / `uv.lock` edits produced while running the script for verification are intentionally not included — they are an intended side effect of external-mode setup on the consumer's working copy, not source changes for this PR.
- No Python/C++ code paths change. The PyPI (`pypi`) and custom-wheel (`apply-custom-rebel.sh`) flows in `dev-setup.sh` are untouched.
- This PR is explicitly scikit-build-core-only. Supporting both layouts was attempted in the first commit but dropped on review — the install step ran unconditionally against legacy trees and would have triggered a fresh CMake/Cython build, invalidating the "legacy supported" claim. The fix-forward is to detect and reject pre-migration trees early.

### DIFF
--- a/tools/build-with-external-rebel.sh
+++ b/tools/build-with-external-rebel.sh
@@ -268,11 +268,16 @@ check_rebel_python_version() {
     local expected_pattern="_C.cpython-${current_py_version}-*.so"
     local rebel_c_so
 
-    # Check in python/rebel first (primary location)
-    rebel_c_so=$(find "$REBEL_HOME/python/rebel" -maxdepth 1 -name "$expected_pattern" 2>/dev/null | head -1)
+    # scikit-build-core places the built extension under python/build/skbuild
+    # (current rebel-compiler layout). Source-tree python/rebel/ only holds .py
+    # files and _C/*.pyi stubs, so look in the build output first.
+    rebel_c_so=$(find "$REBEL_HOME/python/build/skbuild" -maxdepth 1 -name "$expected_pattern" 2>/dev/null | head -1)
 
+    # Legacy layouts (pre scikit-build-core migration) kept _C.so next to __init__.py.
     if [ -z "$rebel_c_so" ]; then
-        # Also check in rebel/python/rebel (alternative location)
+        rebel_c_so=$(find "$REBEL_HOME/python/rebel" -maxdepth 1 -name "$expected_pattern" 2>/dev/null | head -1)
+    fi
+    if [ -z "$rebel_c_so" ]; then
         rebel_c_so=$(find "$REBEL_HOME/rebel/python/rebel" -maxdepth 1 -name "$expected_pattern" 2>/dev/null | head -1)
     fi
 
@@ -288,7 +293,7 @@ check_rebel_python_version() {
 
     # Find all available versions
     local available_versions
-    available_versions=$(find "$REBEL_HOME/python/rebel" "$REBEL_HOME/rebel/python/rebel" -maxdepth 1 -name "_C.cpython-*.so" 2>/dev/null | \
+    available_versions=$(find "$REBEL_HOME/python/build/skbuild" "$REBEL_HOME/python/rebel" "$REBEL_HOME/rebel/python/rebel" -maxdepth 1 -name "_C.cpython-*.so" 2>/dev/null | \
         xargs -I{} basename {} | grep -oP 'cpython-\K\d+' | sort -u)
 
     if [ -n "$available_versions" ]; then
@@ -350,8 +355,11 @@ install_dependencies() {
     # Align with rebel-compiler Quick Install: cmake<4.0, lit
     pip install "cmake>=3.18,<4.0" ninja jinja2 hatchling setuptools-scm editables mypy lit
 
-    # Set environment for rebel-compiler (rebel Python deps installed after uv sync in build_torch_rbln)
-    export PYTHONPATH="$REBEL_HOME/python:$PYTHONPATH"
+    # Native lib paths only. Python import is wired up later by the editable
+    # rebel-compiler install (scikit-build-core), which drops its own redirect
+    # .pth into site-packages — prepending $REBEL_HOME/python to PYTHONPATH here
+    # would let the source-tree rebel/ shadow the editable install and break
+    # `import rebel._C` under the new layout.
     export LD_LIBRARY_PATH="$REBEL_HOME/build:$LD_LIBRARY_PATH"
 
     # Source dynamic_linking.env for additional libraries
@@ -362,11 +370,13 @@ install_dependencies() {
     # Create activate_rebel script
     create_activate_rebel_script
 
-    # Create .pth file for rebel-compiler in site-packages
+    # Remove the stale pre-scikit-build-core rebel_compiler.pth if an older run
+    # of this script left one behind; it would also shadow the editable install.
     local site_packages
     site_packages=$(python -c "import site; print(site.getsitepackages()[0])" 2>/dev/null || echo "")
-    if [ -n "$site_packages" ] && [ -d "$site_packages" ]; then
-        echo "$REBEL_HOME/python" > "$site_packages/rebel_compiler.pth"
+    if [ -n "$site_packages" ] && [ -f "$site_packages/rebel_compiler.pth" ]; then
+        log_info "  Removing stale rebel_compiler.pth from site-packages"
+        rm -f "$site_packages/rebel_compiler.pth"
     fi
 
     log_info "Dependencies installed"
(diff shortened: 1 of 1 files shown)


tools/build-with-external-rebel.sh
```diff
@@ -347,11 +352,17 @@ setup_virtualenv() {
...
     pip install --upgrade pip
-    # Align with rebel-compiler Quick Install: cmake<4.0, lit
+    # Align with rebel-compiler Quick Install: cmake<4.0, lit.
+    # rebel-compiler's own build-system.requires (scikit-build-core, pybind11,
+    # cython, setuptools_scm) are installed later in install_rebel_python_deps,
+    # *after* `uv sync --no-install-project` — otherwise uv sync would wipe them.
     pip install "cmake>=3.18,<4.0" ninja jinja2 hatchling setuptools-scm editables mypy lit
 
-    # Set environment for rebel-compiler (rebel Python deps installed after uv sync in build_torch_rbln)
-    export PYTHONPATH="$REBEL_HOME/python:$PYTHONPATH"
```
rebel-chanheo, line 354: I understand that with the introduction of scikit, `PYTHONPATH` is no longer required when installing rebel-compiler. Is that correct? 

If so, it would be good to organize all the information regarding `PYTHONPATH` mentioned in other files, not just this one, together this time.

rebel-scottlee: Yes, that's right — with the scikit-build-core editable install, `pip install -e` drops `_rebel_compiler_editable.pth` (+ its redirect hook) into the venv's site-packages, so `import rebel` resolves to `$REBEL_HOME/python/build/skbuild/` automatically. There's no need to export `PYTHONPATH` for that anymore, and in fact doing so would let the source tree shadow the editable install and break `import rebel._C` under the new layout.

I went through the rest of the repo and sorted what's left:

| File | Purpose | Related to rebel-compiler install? |
| --- | --- | --- |
| `tools/build-with-external-rebel.sh` | Old export/shim removed in this PR; the 3 remaining mentions are explanatory comments (lines 363, 403, 583) | ✅ Already cleaned up |
| `tools/dev-setup.sh:117-121` | Warns when an inherited `PYTHONPATH` could shadow the editable install | ⚠️ Indirect — acts as a safety net for this same policy |
| `torch_rbln/__init__.py:131,144` | Hints `REBEL_HOME` / `LD_LIBRARY_PATH` / `PYTHONPATH` when `librbln.so` isn't found | ❌ Generic loader diagnostic (`get_dll_directory_candidates` walks `sys.path`) |
| `torch_rbln/diagnose.py:14` | Module docstring | ❌ Description only |
| `torch_rbln/_internal/env_diagnostic.py:4,24,349` | What `python -m torch_rbln.diagnose` dumps + hint when a local repo shadows the wheel | ❌ Generic `sys.path` debugging |
| `test/rbln/test_find_and_load_tvm_library.py:7` | Test docstring asserting that error message | ❌ Describes existing behavior |
| `torch_rbln/csrc/rbln/CMakeLists.txt:25,37` | Injected at build time so stubgen can import the freshly built `_C.so` | ❌ Internal to torch-rbln build, no end-user path |

So the install path is fully PYTHONPATH-free in this PR; the remaining mentions each have their own purpose and aren't really in scope here.

rebel-chanheo (approved):
LGTM

PR #27, opened 2026-04-22
feat(ops): add flash_attention_naive_prefill/decode custom ops

rebel-chanheo:
## Summary of Changes

Register PrivateUse1 implementations for two new unified-KV-cache flash attention custom ops that vllm-rbln's flash-attention backend invokes:

- `rbln_custom_ops::flash_attention_naive_prefill`
- `rbln_custom_ops::flash_attention_naive_decode`

Each follows the existing `paged_attn` pattern — a `torch.compile`-traced `nn.Module` wrapper plus a PrivateUse1 dispatch function wired through the `rbln_custom_ops` IMPL library in `torch_rbln/_internal/register_custom_ops.py`.

Two rbln-vs-CPU reference comparison tests are added in `test/rbln/test_custom_kernel.py`, sharing a `_run_flash_attention_naive` helper. The op signature and CPU reference match vllm-rbln's definition (`vllm_rbln/v1/attention/backends/flash_attention.py`) so the schema is consistent across the two call sites.

---

## Related Issues / Tickets

* Resolves #
* Related to #

---

## Type of Change

- [x] \`feature\` — New capability or functionality

---

## Affected Modules

- [x] Ops/Kernels
- [x] \`torch.compile\`

---

## How to Test

1. \`pytest test/rbln/test_custom_kernel.py -k flash_attention_naive -v\`
2. Both prefill and decode should PASS (rbln output matches CPU reference within the class tolerance).
3. Existing 4 paged_attn tests in the same file remain green.

---

## Checklist

* [x] PR title follows Conventional Commits format
* [ ] This PR is linked to a corresponding issue
* [x] Linting passes
* [ ] All CI tests pass
* [x] The test method is described, and the expected result is clearly stated (if applicable)
* [ ] Relevant documentation has been updated (if applicable)

PR #29, opened 2026-04-22
test(models): add vllm-rbln end-to-end CI suite + install script path

rebel-chanheo:
## Summary of Changes

Adds an end-to-end model-suite test that exercises the vLLM native model path on RBLN device tensors, plus the install-script plumbing so the existing ``python test/run_tests.py --suite=models`` CI entry point brings up vllm + vllm-rbln automatically.

Test matrix (``test/models/test_vllm_llm.py``):

- graph TP1 — Llama-3.2-1B-Instruct, Qwen2.5-1.5B-Instruct, Qwen3-0.6B
- graph TP2 via RSD — Llama-3.2-1B, single vLLM worker over two physical  NPUs through ``VLLM_RBLN_TP_SIZE=2`` (the RCCL multi-worker collective  path is intentionally out of scope here)
- eager TP1 — Llama-3.2-1B with ``VLLM_RBLN_SAMPLER=0`` + ``VLLM_RBLN_COMPILE_MODEL=0``

Each case greedy-decodes five tokens against the prompt "The capital of France is" and asserts on the exact generated string captured during a green run. Drift against those strings is a real correctness regression.

Install pipeline (``tools/test/install-test-deps.sh``):

- vllm is pinned to a CPU wheel from wheels.vllm.ai, version controlled by ``VLLM_VERSION`` (default ``0.18.0+cpu``).
- vllm-rbln is source-cloned from GitHub and installed editable, with configurable ``VLLM_RBLN_REPO`` / ``VLLM_RBLN_REF`` / ``VLLM_RBLN_DIR`` (default branch: ``origin/device_tensor_rebased`` — the branch that introduces ``VLLM_RBLN_USE_DEVICE_TENSOR``).
- ``VLLM_RBLN_SKIP=1`` disables the entire section so environments that cannot install vllm-rbln still complete the rest of the model suite.

The test file itself ``pytest.importorskip``'s ``vllm_rbln``, so even without the install step the suite remains green — the new cases are skipped cleanly.

---

## CI fixes (added 2026-04-23)

Initial CI runs crashed with two separate symptoms. Both are now addressed in this PR.

1. **``RBLN_DEVICES`` snapshot mismatch in EngineCore** (all five cases failed)

   ```
   RuntimeError: RBLNRuntimeError: RBLN_DEVICES environment variable changed at runtime.
   Initial value: , Current value: 0
   ```

   Root cause: rebel's ``librbln-thunk.so`` snapshots ``RBLN_DEVICES`` at first device-handle construction (very early in subprocess startup, before ``RBLNWorker.__init__`` runs). Stock vllm-rbln's ``RBLNWorker._init_device_env`` then re-assigns ``RBLN_DEVICES`` to the expanded ``VLLM_RBLN_TP_SIZE``-based mapping, which rebel flags as a runtime env mutation on the next ``get_npu_name`` call.

   Fix (``test/models/test_vllm_llm.py``, ``test/models/_vllm_rbln_worker_patch.py``):
   - Pre-set ``RBLN_DEVICES`` (``"0"`` for TP1, ``"0,1"`` for TP2) *before* any vllm/vllm_rbln import in ``_vllm_generate_worker`` so rebel's snapshot matches the final value from the start.
   - Also pre-set ``RBLN_NPUS_PER_DEVICE`` for TP>1 (current vllm-rbln no longer sets it, unlike the pre-rewrite version in ``vllm_rbln_bak``).
   - Ship a patched ``RBLNWorker`` subclass that makes ``_init_device_env`` a no-op, in its own module so EngineCore can import it by qualified name.

2. **``worker_cls`` pydantic validation failure** (follow-up of the first fix attempt)

   ```
   pydantic_core._pydantic_core.ValidationError: 1 validation error for ParallelConfig
   worker_cls
     Input should be a valid string, unable to parse raw data as a unicode string
     [type=string_unicode, input_value=b"\x80\x05\x95..."]
   ```

   Root cause: ``LLM.__init__`` cloudpickles class-typed ``worker_cls`` into bytes, but current vLLM's ``ParallelConfig.worker_cls: str`` rejects bytes via pydantic and ``worker_base.py`` explicitly declares "passing worker_cls is no longer supported. Please keep the class in a separate module and pass the qualified name of the class as a string."

   Fix:
   - Move ``PatchedRBLNWorker`` into its own module (``test/models/_vllm_rbln_worker_patch.py``) and pass ``worker_cls="test.models._vllm_rbln_worker_patch.PatchedRBLNWorker"`` as a qualified string.
   - Prepend the repo root to ``PYTHONPATH`` in ``_run_case``'s ``MonkeyPatch.context`` so EngineCore's spawned Python interpreter can resolve the qualified name (``multiprocessing.spawn`` starts a fresh interpreter without CWD on ``sys.path``).

Relevant commits: ``b2d0ece`` (first fix), ``5ee62e3`` (qualified-string rework).

---

## Related Issues / Tickets

* Resolves #
* Related to #

---

## Type of Change

- [x] \`test\` — Test suite addition

---

## Affected Modules

- [x] Build/Tools
- [x] Docs

---

## How to Test

1. Local dry-run of the install path:
   ```
   bash tools/test/install-test-deps.sh --dry-run
   VLLM_RBLN_SKIP=1 bash tools/test/install-test-deps.sh --dry-run
   ```
2. End-to-end (requires physical NPUs):
   ```
   bash tools/test/install-test-deps.sh
   python test/run_tests.py --suite=models
   ```
3. Individual cases:
   ```
   pytest test/models/test_vllm_llm.py -v
   ```
   All five collected cases should PASS with the observed greedy outputs.
4. Opt-out confirmation: ``VLLM_RBLN_SKIP=1 pytest ...`` cleanly skips
   the new cases via ``pytest.importorskip``.

---

## Checklist

* [x] Linting passes
* [ ] All CI tests pass

---

## Notes

- The end-to-end test currently depends on two backend-side fixes that land outside this PR: the rbln ``torch.compile`` backend stripping non-plain ``nn.Parameter`` subclasses before ``jit.trace`` (``rebel/core/torch_compile.py``) and TVM's PyTorch frontend extracting constants via ``.cpu().numpy()`` for any non-CPU device (``tvm/relay/frontend/pytorch.py``). Without those, the test fails loudly inside the rebel compile path rather than silently passing; that is the intended behavior until the backend catches up.
- torch-rbln-side dependencies (PRs #25 and #27) land independently and are not required to merge this one first.


test/models/test_vllm_llm.py
```diff
@@ -0,0 +1,303 @@
...
+
+    ``RBLN_DEVICES`` / ``RBLN_NPUS_PER_DEVICE`` must be set in the *parent*
+    (``_run_case``) before spawn: rebel's ``librbln-thunk.so`` snapshots them
+    at ``import torch``, and spawn re-imports this module before reaching this
+    body. Same pattern as ``test_optimum_llm.py::_run_test_case``.
+    """
+    from vllm import LLM, SamplingParams
+
+    cfg = MODEL_CONFIGS[model_key]
+    tc = unittest.TestCase()
```
rebel-clee: Using `unittest.TestCase()` just to call its assertion methods inside a pytest-collected function is a fairly well-known antipattern — it skips pytest's `assert` rewriting, so failures lose the rich introspection (operand values, diff) that pytest normally surfaces. In a pure pytest context, the canonical form is plain `assert` with the same message argument — for instance, `assert len(outputs) == 1, "expected a single RequestOutput"` in place of the `tc.assertEqual(...)` call, and the same shape across the other `assertGreater` / `assertTrue` lines.

Happy to leave it as-is if there's context I'm missing on why `unittest`-style asserts are preferred here.

test/models/test_vllm_llm.py
```diff
@@ -0,0 +1,303 @@
...
+        # Native vLLM model path config — matches the vllm-rbln team's
+        # ``test_llama_batch.py`` reference on ``device_tensor_rebased``.
+        mp.setenv("VLLM_USE_V1", "1")
+        mp.setenv("VLLM_RBLN_USE_VLLM_MODEL", "1")
+        mp.setenv("VLLM_RBLN_SAMPLER", "1")
+        mp.setenv("VLLM_RBLN_COMPILE_STRICT_MODE", "1")
+        mp.setenv("VLLM_DISABLE_COMPILE_CACHE", "1")
+        mp.setenv("RBLN_KERNEL_MODE", "triton")
+        mp.setenv("RBLN_PV_OPT", "1")
+        mp.setenv("TORCH_RBLN_DISABLE_FALLBACK", "compile_error")
```
rebel-clee: `_run_case` sets `TORCH_RBLN_DISABLE_FALLBACK="compile_error"` explicitly, but `test/conftest.py::disable_compile_error_fallback` is autouse and already arranges for `compile_error` to be present. The explicit set both shadows the fixture's append-style behavior and silently drops any other categories an external caller might have configured. If the intent here is to narrow it to exactly `compile_error` for the spawned worker, a one-line comment saying so would help the next reader avoid assuming the explicit set is just redundant; if it isn't intentional, the line can probably just go.

tools/test/install-test-deps.sh
```diff
@@ -1,160 +1,181 @@
...
+#
+# transformers and optimum-rbln are NOT installed here — they arrive
+# transitively via vllm-rbln in step 4. Environments that opt out of vllm-rbln
+# (VLLM_RBLN_SKIP=1) and still want to run those tests must install both
+# manually.
+
+install_model_test_deps() {
+  log_step "Model-test deps (torchvision CPU + pandas)"
+  pip_install "torchvision==0.25.0+cpu" \
+    --index-url https://download.pytorch.org/whl/cpu
```
rebel-clee: One thing worth flagging — the old flow used to verify that `torchvision` came from `https://download.pytorch.org/whl/cpu` and would refuse to proceed if the source or version didn't match. The new flow just calls `pip install torchvision==0.25.0+cpu --index-url <cpu-index>` and trusts pip's resolver.

If a developer already has a non-CPU `torchvision` wheel installed (the cross-poison case the original script was guarding against), pip will skip the reinstall, and the conflict only surfaces later as a runtime operator-registration error deep inside a test. Might be worth keeping a minimal version + source check here, or at least adding `--force-reinstall` to the `torchvision` install.

tools/test/install-test-deps.sh
```diff
@@ -1,160 +1,181 @@
...
+        sys.exit(0)
+sys.exit(f"vllm-cpu index URL not found in {sys.argv[1]}")
+PY
+}
+
+install_vllm_rbln() {
+  if [ "${VLLM_RBLN_SKIP:-0}" = "1" ]; then
+    log_step "vllm-rbln (skipped via VLLM_RBLN_SKIP=1)"
+    return 0
+  fi
```
rebel-clee: The model test suite is meant to validate the vllm-rbln runtime path. `test_vllm_llm.py` is built around it, and with this PR, `test_optimum_llm.py` and `test_transformers.py` also end up depending on vllm-rbln — the script no longer installs `transformers` / `optimum-rbln` standalone, so those wheels come in only as transitive deps through vllm-rbln. vllm-rbln is the whole reason the suite exists, not an optional add-on.

So `VLLM_RBLN_SKIP=1` ends up framing `--suite=models` as "model tests minus vllm-rbln when convenient," which I think weakens the signal in both directions:

  * with the skip applied, `test_vllm_llm.py` `importorskip`s cleanly — every case silently skips, the suite goes green, and we get zero coverage on the exact path the file is meant to exercise;
  * `test_optimum_llm.py` and `test_transformers.py` get collected and then fail with `ImportError`, since neither `transformers` nor `optimum-rbln` is installed when vllm-rbln is skipped.

Would it be reasonable to treat vllm-rbln as a hard dependency of `--suite=models` instead? Concretely:

  1. drop the `VLLM_RBLN_SKIP=1` early-return in `install_vllm_rbln`;
  2. drop the `VLLM_RBLN_SKIP=1` line from the script-header env-var docs;
  3. swap `pytest.importorskip("vllm_rbln")` in `test_vllm_llm.py` for a hard `import vllm_rbln`, so the suite fails loudly at collection time when vllm-rbln is missing, instead of silently skipping every case.

test/models/test_vllm_llm.py
rebel-chanheo: Addressed. Replaced the unittest.TestCase() assertions with plain pytest assert statements.

test/models/test_vllm_llm.py
rebel-chanheo: Addressed. Removed the explicit TORCH_RBLN_DISABLE_FALLBACK override and now rely on the autouse fixture to preserve append-style behavior.

tools/test/install-test-deps.sh
rebel-chanheo: Addressed. Added --force-reinstall --no-deps to the torchvision CPU wheel install so an existing non-CPU wheel does not get reused.

tools/test/install-test-deps.sh
rebel-chanheo: Addressed. Removed VLLM_RBLN_SKIP and changed test_vllm_llm.py to hard-import vllm_rbln so the model suite fails loudly when vllm-rbln is missing.

rebel-clee (approved):
LGTM

PR #46, opened 2026-05-11
perf(eager): dispatch shim + view-on-device + fp16 quirks + perf recovery

rebel-chanheo:
## Summary of Changes

Foundation for eager-mode LLaMA-class workloads on RBLN: a C++ dispatch shim with a
per-runtime warm cache (3.9× hit-path speedup vs Python wrapper), a generic
view-on-device detector that emits view recipes inside the compiled FX graph, a
v-memory borrow path through `cpu_fallback_rbln` (no D2H/H2D round-trip for
`out=` slots), and the dispatch / fp16 correctness fixes those changes require
to keep `test_set_ci` green. Net effect:

- LLaMA-1B eager prefill 12.92 s → 8.91 s (-31%) via BFS-key view-recipe cache
- LLaMA-1B eager decode rotary_rotate_half -98% via 64-elem align-fallback
- Warm-cache hit path 3.9× faster than the Python wrapper micro-bench
- LLaMA-8B graph 155 → 193 tok/s

## Related Issues / Tickets

* Related to #

## Type of Change

- [x] `core` — RBLN backend dispatch + view-on-device + warm cache
- [x] `perf` — eager-mode latency recovery

## Affected Modules

- [x] Backend
- [x] Ops/Kernels
- [x] Build/Tools (codegen template)

## How to Test

```bash
python test/run_tests.py --suite=core --suite=ops
python test/run_tests.py --suite=ops --test_mode=release
```

`lintrunner --merge-base-with origin/dev` clean. 1627 ops tests pass (release mode), 0 failed.

## Notes

Stacked: PR #55 (`chan/perf-foundation-native-ops`) builds on this branch and lands
fill_/arange/_local_scalar_dense/clone native impls on top.


rebel-clee:
I appreciate the work here — the performance gains are substantial, and the warm-cache and self-registering fast-path approach was easy to follow.

More broadly, the in-code comments carry a lot of valuable context — the dated reproduction notes especially — but several run long enough to bury the code they annotate. It might be worth tightening the longest ones so the logic stays in the foreground.

I've left inline comments throughout — mostly questions to confirm intent, plus a few concrete changes I'd like to get in. Happy to discuss any of them.

torch_rbln/_internal/ops_utils.py
```diff
@@ -313,6 +440,1426 @@ def prepare_args_for_contiguous(args, kwargs_filtered):
...
+
+def view_recipes_for_positional(view_recipes_flat, positional_count):
+    """Slice the flat ``view_recipes`` tuple from ``prepare_args_view_aware``
+    down to just the leading ``positional_count`` entries (matching the
+    positional args of an OpModule.forward).
+    """
+    return view_recipes_flat[:positional_count]
+
+
+def compile_and_run_view_aware(op_callable, op_name, args, kwargs_filtered, out_tensor):
```
rebel-clee, line 1656: The per-op compiler workarounds in `compile_and_run_view_aware` are accumulating as inline `if op_name == ...` branches that increasingly bury the core compile-and-run flow. Would it be worth lifting these into a small table keyed by op name (like the fast-path registry)? That would keep the main path clean and each workaround self-contained and testable.

torch_rbln/_internal/ops_utils.py
```diff
@@ -313,6 +440,1426 @@ def prepare_args_for_contiguous(args, kwargs_filtered):
...
+#
+# LLaMA-1B prefill exercises ~48 BFS calls across 16 layers × ~3 view
+# patterns. The data_ptr-keyed cache previously prototyped never hit
+# (per-layer storages differ); a geometric key hits 16x for each of the
+# ~3 unique patterns. Saves ~4 s of the ~12 s prefill (measured 2026-05-11).
+#
+# Sentinel ``_RECIPE_CACHE_NONE`` marks "no recipe found" so the negative
+# result is cached too (BFS exhaustion is the expensive case).
+_RECIPE_CACHE_NONE = object()
+_recipe_cache: dict = {}
```
rebel-clee, line 1275: `_recipe_cache` has no eviction — `view_recipe_cache_reset()` is benchmark-only and `empty_cache()` skips it — so under a variable-shape serving workload it grows unbounded for the life of the process. It's also updated without a lock, unlike the sibling `compile_cache`, which uses a `threading.Lock`; same for the `_hits` and `_misses` counters. Could we cap it (or have `empty_cache()` clear it) and match the compile cache's locking?

aten/src/ATen/native/rbln/fast_paths/mean_lastdim_fp32.cpp
```diff
@@ -0,0 +1,92 @@
...
+    return false;
+  }
+  const int64_t inner = self.size(self_dim - 1);
+  if (inner <= 0) {
+    return false;
+  }
+  const int64_t outer = self.numel() / inner;
+  if (out.numel() != outer) {
+    return false;
+  }
```
rebel-clee: The `rsqrt` and `pow` handlers pin the full output shape (`out.sizes() != self.sizes()`), but `mean_lastdim_handler` only checks `out.numel() == outer`. I get that it's a reduction — with `keepdim=True` the result is `[..., 1]`, so the sibling check wouldn't apply as-is — but an `out=` tensor with the right element count and a different shape still passes the guard and comes back un-resized, where core `mean.out` would have resized it. Was the numel-only check deliberate, or should it also pin the keepdim shape? Separate and optional: the three handlers share the same dtype and contiguity checks, so pulling those into a helper (each keeping its own shape check) might be worth it as the set grows.

aten/src/ATen/native/rbln/RBLNCPUFallback.cpp
```diff
@@ -96,6 +190,74 @@ bool validate_tensor_list(const c10::List<at::Tensor>& tensorlist) {
...
+    for (size_t k = 0; k < filled.size(); ++k) {
+      cpu_tensors[leftover_indices[k]] = std::move(filled[k]);
+    }
+  }
+  return cpu_tensors;
+}
+
 } // namespace
 
 void cpu_fallback_rbln(
```
rebel-clee, line 272: Rejection handling in `cpu_fallback_rbln` looks asymmetric: the read-borrow sites use `try_borrow_host_ptr` and fall back to a device-to-host copy when the runtime rejects a borrow, but the pure-`out=` branch's `acquire_host_ptr_for_overwrite` throws instead. Could an `out=` tensor hit that rejection too — and if so, should it fall back the same way (the `at::empty` branch is right below it)? Separately, a contiguous offset view (`x[1:]`) passes the `is_contiguous()` guard despite a non-zero `storage_offset()` — for an accepted borrow, is `host_ptr` resolved at the offset, or at the base allocation?

torch_rbln/_internal/ops_utils.py
rebel-chanheo: Agreed on the readability — the guards do crowd the top of `compile_and_run_view_aware`.

One thing that makes a plain op-name table awkward: the three workarounds key on different things and do different things.
- the 64-align fallback keys on *tensor shape*, not op name, and early-returns `cpu_fallback_path`;
- the `div` trunc/floor case keys on op name **+ a kwarg + dtype**;
- the comparison-ops case keys on an op set but doesn't short-circuit — it *rewrites the args* to `.contiguous()` and falls through to the normal path.

So an `op_name -> handler` dict wouldn't capture the shape-based one and would have to encode "short-circuit vs. rewrite-and-continue" anyway. If we lift them, I think the right shape is an ordered list of pre-dispatch hooks with a common signature — `(op_name, args, kwargs) -> Optional[result] | (args, kwargs)` — where returning a value short-circuits and returning args replaces them. That keeps the main flow to a couple of hook calls and makes each workaround independently testable, which is the real win you're after.

Since there are exactly three today and each is self-contained, I'd lean toward extracting each into a named helper now (so the main path reads as `if (r := _maybe_fallback_unaligned(...)): return r`) and deferring the hook registry until the set actually grows — happy to do the helper extraction in this PR if you'd prefer it landed here rather than as a follow-up.

torch_rbln/_internal/ops_utils.py
rebel-chanheo: Addressed in 1259bad: `empty_cache()` now also calls `view_recipe_cache_reset()`, so the cache is bounded. It's metadata-only (no device buffers) so it never showed up in `memory_stats`, but clearing it here keeps "let go of everything the user isn't holding" complete. (I left the sibling `_compiled_op_cache` as-is, since `empty_cache()` doesn't clear that one either — happy to wire both up if you'd prefer the symmetry.)

On the lock: I deliberately kept the recipe cache lock-free. The access is pure-Python dict get/set under the GIL, and the recipe is a function of metadata alone, so a race can at worst recompute one BFS (idempotent) or drop a `_hits`/`_misses` increment (diagnostics only) — no corruption. The compile cache's lock guards a heavier critical section (compile + insert), which is why it has one. Let me know if you'd still like them matched for consistency.

aten/src/ATen/native/rbln/fast_paths/mean_lastdim_fp32.cpp
rebel-chanheo: Good catch — addressed in 1259bad. The numel-only check wasn't deliberate. The handler now pins the full keepdim shape (`self.sizes()` with the reduced last dim set to 1), matching the rsqrt/pow siblings, and bails to core `mean.out` for any other shape so it gets resized as eager would.

aten/src/ATen/native/rbln/RBLNCPUFallback.cpp
rebel-chanheo: Both addressed in 1259bad.

- **Borrow symmetry:** added `try_acquire_host_ptr_for_overwrite` (non-throwing, mirrors `try_borrow_host_ptr`). The `out=` overwrite borrow now falls back to a fresh `at::empty` + writeback on a runtime rejection instead of throwing. The Step 3 write-back was the other half of the asymmetry — it re-called the throwing `acquire` — so I converted that to the same `try` + `_copy_from_and_resize` fallback, making the whole `out=` path throw-free.
- **Offset view:** to your question — `data_ptr()` on `x[1:]` is an interior vaddr, and the runtime borrow resolves against the backing allocation, not the offset, so the borrowed host view would not be offset-correct. I guarded `storage_offset() != 0` out of the borrow path on both the read and `out=` sides (copy path instead). This is the conservative choice; if we later confirm the runtime offset-resolves interior vaddrs, the guard can be relaxed for the perf win.

torch_rbln/_internal/ops_utils.py
rebel-clee: Thanks — the `empty_cache()` hook looks good, and fair point on the lock: not worth the per-op cost on the hot path.

torch_rbln/_internal/ops_utils.py
rebel-clee: Agreed — the hook list is the right shape. Happy to take it as a follow-up rather than grow this PR.

rebel-clee (approved):
LGTM

PR #47, opened 2026-05-11
perf(eager): view-on-device + fp16 fixes + 64-elem align-fallback

rebel-chanheo:
## Summary

Lands the view-on-device generic detector, the fp16 quirk handling discovered while validating it, the align-fallback that recovers the LLaMA-1B rotary regression introduced by view-on-device, and a BFS shape-key cache that brings prefill below the pre-regression sweet spot. Stacked on top of #46 (V4 foundation).

This is PR 2 of 3 stacked PRs:
- **#46** Foundation — V4 shim + warm cache + borrow + native factory + cpu_fallback perf
- **#47 (this PR)** View-on-device + fp16 quirks + align-fallback + BFS geometric-key cache + custom float tolerance
- **#48** Bench harness + c10::rbln borrow unify + constraint pin to rebel-compiler `045b3ed0`

## BFS geometric-key cache (added on top of align-fallback)

Zero-base profile on the post-align-fallback HEAD identified `_bfs_search_recipe` as the dominant prefill cost: 48 BFS calls at ~93 ms cumulative each = **4.5 s of 12.9 s prefill (33%)**. The detector's output is a function of metadata alone (data_ptr / Tensor._version do not affect the search), so a geometric-only cache key `(base.size, base.stride, t.size, t.stride, t.storage_offset)` is sound.

A prior G1 prototype keyed by `(data_ptr, _version, size, stride, offset)` never hit across layers (data_ptr differs per storage); the geometric key collapses LLaMA-1B's ~3 unique view patterns to 3 cache entries that 16 layers all hit.

Measured on rebel-compiler 045b3ed, LLaMA-1B prompt=32 / decode=8:
- Prefill wall:   12.92 s → **8.91 s**  (**-31%**)
- Decode wall:    360.8 → 355.0 ms (-1.6%, within noise)
- BFS calls:      48 → 3 (-94%)
- Cache hit rate: **93.8%** (45 hits / 3 misses / 3 unique keys)

## View-on-device regression and its fix

`feat(view-on-device): generic detector + simulation + blacklist + cache key fixes` introduces a BFS-based view-recipe detector that emits FX `aten::expand/permute/narrow` ops so the view materialization runs on-device. It is a significant feature for graph-mode workloads but causes a measured **+155% / +183%** prefill/decode regression on LLaMA-1B eager mode — `rotary_rotate_half` accounts for 87-89% of the regression mass.

Mechanism (root cause confirmed by IR dump):

```
main fn: cast(f16→custom float) → device_fn(strided_slice+negative) → depad(-32) → cast(custom float→f16)
```

The `strided_slice` device-lowers fine, but the surrounding `contrib_dummy_cast` (f16↔custom float) and `contrib_aligned_pad(-32)` (64-align depad) lack RTOSA device-lowering and always go through the host transform path.

`perf(view-on-device): route 64-elem-unaligned single-tensor ops through cpu_fallback` routes single-tensor ops whose last-dim element count is not a multiple of 64 through `cpu_fallback_path` instead. Recovers rotary_rotate_half decode by **-98%** without disabling view-on-device for the workloads where it pays off.

## fp16 quirks

- `fix(view-on-device): reject 0-dim base in view recipe detector` — autograd bprop's `scalar.expand(N)` triggered rebel-compiler `build_internal` abort on fp16.
- `fix(view-on-device): materialize view inputs for comparison ops` — eq/ne/lt/gt on view inputs hit the same IR abort pattern.
- `fix(div): route fp16 div trunc/floor through cpu_fallback` — rebel-compiler host-fused trunc_divide misalignment; reproducer at `docs/bug_reports/rebel_host_fused_trunc_divide_wrong_rows.py`.
- `fix(test_out): work around RBLN fp16 -inf eq/sub quirk via CPU compare` — fp16 `-inf` comparison via `eq/sub` returns wrong polarity on device.
- `test(rbln): root-cause notes for fp16 skips (view-on-device IR aborts)` — three remaining narrowly-scoped skips with documented leads.
- `test(view-on-device): custom float-aware fp16 tolerance for non-zero offset tests` — view-on-device's cast(f16↔custom float) round-trip introduces ~1 ULP drift per elementwise op; multi-stage view chains in `test_non_zero_storage_offset.py` accumulate the drift and trip default-tight tolerance. Adds the same custom float tolerance pattern that 30584d2 added for `test_ops.py`, scoped at the base TestCase class. Result: **10 failed → 0 failed**.

## Correctness fixes

- `fix(cpu_fallback): classify Tensor? args in schema-typed arg_kind dispatch` — `OptionalTensor` kind missing from V4 shim's schema-typed dispatch.
- `fix(warm-cache): bypass on non-contiguous out= dispatch` + `fix(rbln): correct out= shape handling on warm-cache and cpu-fallback paths`.

## Measurements

LLaMA-1B eager (n=5):

| Stage   | Pre-`73e84fc` (`ca82126`) | Post-`73e84fc` + previous fixes | + align-fallback | **+ BFS cache (this PR final)** |
|---------|--------------------------:|-------------------------------:|---------------------------:|---------------------------:|
| Prefill | 9180 ms                   | 24428 ms (+166%)               | 12410 ms (-49%)            | **8906 ms** (-3% vs `ca82126`) |
| Decode  | 446 ms                    | 1324 ms (+197%)                | 403 ms (-69%)              | **355 ms** (-21% vs `ca82126`) |
| rotary_rotate_half decode | 70 ms             | 862 ms (+1131%)                | 13 ms (-98%)               | 13 ms |

The new BFS-cache commit brings prefill below the `ca82126` sweet spot. The remaining gap to graph mode (eager 8.9 s / graph 34 ms) lives in the rebel-runtime `EnsureSyncedOnPhysicalView` path (prep_in/prep_out 99.5-99.8% of warm-hit cost) and is tracked as R1-R4 escalation on the rebel-compiler side.

## Test status

- `test/rbln/` + `test/internal/` smoke (1199 tests including the storage-offset suite the tolerance commit fixes): **1199 pass, 1 skip, 0 fail**.
- A prior version of this PR description incorrectly described 10 `test_non_zero_storage_offset.py` failures as pre-existing. Re-verified on `chan/perf-foundation` (PR #46 base): those tests pass there. The failures were introduced by `73e84fc` view-on-device and are now fixed by the custom float-aware tolerance commit added in this PR.
- Other broad-run failures (fp16 `test_op_caching`, `test_eager_matches_graph`, c10d_rbln_ccl) ARE pre-existing on `rblnsw/dev` HEAD — verified directly.

## Risk + revert plan

- BFS cache revert: pure Python rollback, no ABI impact.
- Align-fallback revert: 23-line `ops_utils.py` rollback.
- View-on-device revert: rolls back to `chan/perf-foundation` foundation.
- Tolerance fix revert: re-introduces the 10 test_non_zero_storage_offset failures but no production regression.

### DIFF
--- a/torch_rbln/csrc/rbln/DispatchShim.cpp
+++ b/torch_rbln/csrc/rbln/DispatchShim.cpp
@@ -125,9 +125,15 @@ struct SchemaCache {
   bool populated = false;
   bool is_align_sensitive = false; // op name in align_sensitive_ops() set
   bool is_broadcast_op = false; // op may broadcast tensor args (mul/add/sub/...).
-                                 // When true, build_cache_key uses post-broadcast
-                                 // shapes and try_warmcache_hit broadcasts inputs
-                                 // before passing data_ptrs to PrepareInputs.
+                                 // When true, ``try_warmcache_hit`` decides
+                                 // per-call (via ``needs_last_dim_one_broadcast``)
+                                 // whether to materialize the post-broadcast
+                                 // contig buffers before passing data_ptrs to
+                                 // PrepareInputs. ``build_cache_key`` always
+                                 // uses RAW input shapes regardless — keying on
+                                 // post-broadcast shapes earlier caused
+                                 // (4,8,16)+() and (4,8,16)+(4,8,1) to collide
+                                 // on a single key (2026-04-30).
 };
 
 struct ShimEntry {
@@ -511,44 +517,6 @@ inline bool all_input_shapes_equal(torch::jit::Stack* stack, const SchemaCache& 
     }
   }
   return true;
-}
-
-// For broadcast ops, compute the broadcast shapes that the runtime was
-// compiled for. For non-broadcast ops, returns empty (caller uses raw shapes).
-// On any error (mismatched shapes, etc.) returns empty too — caller falls
-// back to raw shapes which will likely runtime_failed and erase, but that's
-// no worse than current behavior.
-std::vector<std::vector<int64_t>> compute_broadcast_shapes(
-    torch::jit::Stack* stack, const SchemaCache& cache) {
-  std::vector<std::vector<int64_t>> result;
-  if (!cache.is_broadcast_op) return result;
-  // Fast-path: all tensor inputs already same shape → no broadcast needed,
-  // raw shapes are correct keys. Skip the expensive at::broadcast_tensors call.
-  if (all_input_shapes_equal(stack, cache)) return result;
-  auto arguments = torch::jit::last(stack, cache.num_args);
-  std::vector<at::Tensor> tensor_args;
-  std::vector<size_t> tensor_arg_indices;
-  for (size_t i = 0; i < cache.num_args; ++i) {
-    const auto& iv = arguments[i];
-    if (!iv.isTensor()) continue;
-    const auto& t = iv.toTensor();
-    if (!t.defined()) continue;
-    if (cache.is_write_alias[i]) continue;
-    tensor_args.push_back(t);
-    tensor_arg_indices.push_back(i);
-  }
-  if (tensor_args.size() < 2) return result;  // no broadcast needed
-  std::vector<at::Tensor> broadcasted;
-  try {
-    broadcasted = at::broadcast_tensors(tensor_args);
-  } catch (...) {
-    return result;
-  }
-  result.reserve(broadcasted.size());
-  for (const auto& bt : broadcasted) {
-    result.emplace_back(bt.sizes().begin(), bt.sizes().end());
-  }
-  return result;
 }
 
 // Build a WarmCache::CacheKey from the current stack's last num_args IValues.


rebel-chanheo:
⚠️ This PR was **auto-marked merged by GitHub** when `chan/perf-foundation` was force-pushed to include this PR's original head commit (`dfa5799`) in its history — not via the merge button.

### What actually happened

CI on the standalone #46 head was failing on 4 core tests + 14 ops tests because the warm-cache + dispatch-shim + cpu_fallback foundation on #46 needs the view-on-device + Optional[Tensor] dispatch + non-contig out= + fp16 div/eq quirks fixes that originally lived here. To unblock #46, the bulk of this PR's commits were moved onto `chan/perf-foundation`:

- The 16 fix + test commits (view-on-device detector, Optional[Tensor] arg_kind, warm-cache non-contig out, fp16 div/eq quirks, etc.) are now on **PR #46** as untouched commit SHAs.
- The 3 perf-only commits (`b134ac3` 64-elem align-fallback, `8b33e37` BFS geometric-key cache, `0fa6593` non-zero offset fp16 tolerance test) are split into **PR #52** stacked on top of the new `chan/perf-foundation`.

### Verification

- Tree of new `chan/perf-foundation` = original `dfa5799` (this PR's head) **minus** the 3 perf commits' diffs.
- 3 explicit `Revert` commits sit at the tail of new `chan/perf-foundation` so the change is auditable.
- PR #52 cherry-picks the 3 perf commits back on top with `--allow-empty` semantics; tree of PR #52 head = original `dfa5799` (bit-identical, including formatting).

### Net effect for reviewers

- **PR #46**: foundation + view-on-device + fp16 fixes → CI-stable on its own.
- **PR #52**: 3 perf optimizations only — easy to review independently.
- This PR (#47): superseded; nothing to do here.

PR #49, opened 2026-05-13
feat(aten): native v2v cat / index_select / index_copy / repeat_interleave kernels

rebel-chanheo:
## Summary of Changes

Native v2v (device-to-device) RBLN kernels for the cat / index_select / index_copy / repeat_interleave family — outputs are built directly between RBLN tensors via `c10::rbln::memcpy_v2v`, replacing the CPU fallback. `aten::stack` inherits the cat path through `CompositeImplicitAutograd`. `index_copy_` / `index_copy` and `repeat_interleave.self_{int,Tensor}` inherit their respective paths through structured-delegate / composite dispatch.

All ops share a single engine (`at::native::rbln::strided_v2v_copy`) and a backend-API abstraction (`c10::rbln::V2VBatch`). When rebel exposes batched / strided `memcpy_v2v`, only V2VBatch internals change — kernels using the engine pick up the benefit without source edits.

### Per-op notes

**index_copy.** Two-phase: (1) initialise `out` with `self`'s contents via the engine, skipped when `out` aliases `self` (the in-place `index_copy_` dispatch); (2) for each consecutive index run, slice `source` and `out` along `dim` and hand the views to the engine.

**repeat_interleave.** Register the `.Tensor` primitive (the index-builder); the `.self_int` and `.self_Tensor` forms are CompositeImplicitAutograd upstream and decompose into `.Tensor` + `index_select`, so the whole composite chain runs without a host round-trip of `self`.

### File layout

Op kernels live in files that mirror their upstream `aten/src/ATen/native/` location:

```
RBLNTensorShape.{h,cpp}              ← TensorShape.cpp        (cat, stack via composite)
RBLNTensorAdvancedIndexing.{h,cpp}   ← TensorAdvancedIndexing.cpp (index_select, index_copy)
RBLNRepeat.{h,cpp}                   ← Repeat.cpp             (repeat_interleave)
RBLNStridedV2V.{h,cpp}               ← (RBLN-only) strided copy engine
RBLNIndexUtils.h                     ← (RBLN-only) host index helpers shared by indexing ops
c10/rbln/RBLNV2VBatch.{h,cpp}        ← (RBLN-only) backend-API abstraction
```

New ops in the same upstream files should land in the same RBLN file (e.g. future `index_add` / `gather` go in `RBLNTensorAdvancedIndexing.cpp`).

### Opinfo conformance fixes (cat / index_select / stack)

0-D `self` for `index_select`, run-coalescing gated on stride-at-axis (was reading padding NaN from non-contig `self`), non-contig `out` staging with the canonical resize `UserWarning`.

### Always-v2v `RBLNCopy.cpp` (attempted, reverted)

A copy-side always-v2v refactor was attempted alongside this work and **reverted**: per-call `rbln_memcpy_v2v` overhead times outer-iteration count K dominates the saved PCIe round-trip for transpose / permute / inner-dim broadcast / column-slice patterns. `RBLNCopy.cpp` continues using the existing host staging path until rebel exposes batched / strided primitives.

---

## Related Issues / Tickets

* Resolves #

---

## Type of Change

- [ ] `fix` — Bug fix

---

## Affected Modules

- [ ] Backend
- [ ] Device
- [x] Ops/Kernels
- [ ] `torch.compile`
- [ ] Memory
- [ ] Build/Tools
- [ ] Docs

---

## How to Test

```bash
pytest test/python/test_copy_v2v.py \
       test/python/test_cat_index_select_v2v.py \
       test/python/test_index_copy_v2v.py \
       test/python/test_repeat_interleave_v2v.py
pytest test/ops/test_ops.py -k 'cat or index_select or index_copy or repeat_interleave or stack'
build/bin/test_cpp_RBLNV2VBatchTest
```

Expected: all green. The v2v suites cover engine semantics, the four kernels, and stride / broadcast / aliasing / 0-D / 0-numel / in-place edge cases. The opinfo run confirms conformance. The C++ test exercises the V2VBatch contract directly.

---

## Checklist


aten/src/ATen/native/rbln/RBLNCat.cpp
```diff
@@ -0,0 +1,253 @@
+#include <ATen/native/rbln/RBLNCat.h>
```
rebel-clee: The new files break the `RBLN<UpstreamFile>` mirroring convention the rest of this directory follows:

| RBLN file                       | upstream PyTorch file                             |
|---------------------------------|---------------------------------------------------|
| `RBLNCPUFallback.cpp`           | `aten/src/ATen/native/CPUFallback.cpp`            |
| `RBLNCopy.cpp`                  | `aten/src/ATen/native/Copy.cpp`                   |
| `RBLNResize.cpp`                | `aten/src/ATen/native/Resize.cpp`                 |
| `RBLNTensorFactories.cpp`       | `aten/src/ATen/native/TensorFactories.cpp`        |
| `RBLNCat.cpp` (this PR)         | `aten/src/ATen/native/TensorShape.cpp`            |
| `RBLNIndexSelect.cpp` (this PR) | `aten/src/ATen/native/TensorAdvancedIndexing.cpp` |

CUDA / MPS / XPU backends do the same upstream-mirror (`cuda/TensorShape.cu`, `cuda/Indexing.cu`, `mps/Shape.mm`). Renaming to `RBLNTensorShape.{cpp,h}` and `RBLNTensorAdvancedIndexing.{cpp,h}` gets two follow-on wins:

  - `stack` / `chunk` / `unbind` / `split` / `narrow` land naturally in `RBLNTensorShape.cpp` when they get native kernels; same shape for `index_add` / `index_put` / `gather` in `RBLNTensorAdvancedIndexing.cpp`. Single-op files become a proliferation problem.
  - `git grep "::cat"` against upstream lands a reader in `TensorShape.cpp`, and the RBLN-side file matches at a glance.

Could we rename these before merge? Doing it now saves follow-up PRs from either stacking new ops on the wrong-shaped file or paying a churn-rename cost later.

rebel-chanheo: Good point. I have taken it into account.

test/python/test_cat_index_select_v2v.py
```diff
@@ -0,0 +1,641 @@
+# Owner(s): ["module: PrivateUse1"]
```
rebel-clee: For consistency with the rest of `test/rbln/`, could we bring the four new files under `test/python/` in line with `docs/TEST_GUIDE.md`? The guide's Minimal Test Template is what every existing `test/rbln/*` file follows, and the four new files diverge from it in the same five spots:

  - marker: no `@pytest.mark.test_set_ci`. Per the CI/CD Integration section, CI runs `pytest -m "test_set_ci"`, so the 360+ cases here — the v2v-specific stride, non-contig, and in-place scenarios that have no other coverage — are never selected;
  - location: the guide places RBLN-side tests under `test/rbln/`; `test/python/` is new and unused by any existing test;
  - structure: the Minimal Test Template uses `class TestXxx(TestCase)` + `instantiate_device_type_tests(..., only_for="privateuse1")` + `run_tests()`, whereas these files use free pytest functions + `pytest.main()`;
  - env: `test/utils_v2v.py` collects the env-var setup in one place, but `os.environ.setdefault` still bypasses the `enable_deploy_mode` / `enable_eager_malloc` fixtures the guide documents in its Global Fixtures table — those use `monkeypatch.setenv` and revert per test, so env changes don't carry over between tests;
  - duplication: `test/rbln/test_registered_ops.py` already has a `test_cat` method, and the new files add overlapping coverage. Folding them into the existing file (or one consolidated file in `test/rbln/`) lines them up with the established convention.

Following the Minimal Test Template wholesale closes all of these in one go and keeps the new ops consistent with the rest of `test/rbln/`.

aten/src/ATen/native/rbln/RBLNStridedV2V.cpp
```diff
@@ -0,0 +1,114 @@
...
+      inner_block_bytes,
+      c10::IntArrayRef(outer_sizes_vec),
+      c10::IntArrayRef(src_byte_strides),
+      c10::IntArrayRef(dst_byte_strides));
+}
+
+void strided_v2v_copy(const at::Tensor& dst, const at::Tensor& src) {
+  c10::rbln::V2VBatch batch;
+  strided_v2v_copy(dst, src, batch);
+  // batch.submit() runs on destructor.
```
rebel-seungchul: I think the current "RAII flush by default, explicit `submit()` optional" model should be flipped to **"explicit 
  `submit()` required on the success path, destructor is a leak-prevention safety net only."** A few concrete reasons:

  ### 1. The destructor can throw — the header already admits this is a hazard

  From `RBLNV2VBatch.h`:

  > The destructor's submit is best-effort and will RBLN_CHECK-fail loudly if the backend rejects a queued op — call submit() explicitly if you need to handle errors outside of 
  destruction.

  If `RBLN_CHECK` is throw-based (TORCH_CHECK family), a backend rejection during stack unwinding turns into `std::terminate`. The implementation comment ("same hazard pattern as any 
  RAII flush") understates this — `std::fstream::~fstream()` swallows errors precisely to avoid this footgun, and that guard isn't present here.

  ### 2. Flush is tied to object lifetime, not to control flow

  ```cpp
  c10::rbln::V2VBatch batch;
  // ... enqueue ...
  return out;   // ← actual v2v happens here
  ```

  - Locals are destroyed in reverse construction order. If anyone later introduces a staging tensor after batch, the staging buffer is destroyed first — and if it's the source of a
  queued copy, we get a use-after-free. The current call sites are accidentally safe (sources are caller-owned self/out), but the API doesn't enforce that invariant for future
  refactors.
  - Any synchronous memcpy_v2h / synchronize() issued before batch goes out of scope races against the un-submitted queue. This is a very hard bug to spot at the call site.

### 3. Every call site repeats // batch.submit() runs on destructor.

  RBLNStridedV2V.cpp:105, RBLNTensorAdvancedIndexing.cpp:101, RBLNTensorShape.cpp:178 — nearly every user of V2VBatch annotates the implicit behavior. The whole point of RAII is "you
  don't have to think about it." If callers consistently need to comment the implicit flush, the abstraction is leaking and an explicit submit() would be clearer.

### 4. Errors collapse onto the closing brace

  With explicit submit(), a backend rejection points at the submission line and the preceding enqueues are obvious context. With destructor submit, the throw fires at function exit —
  callers see "this function blew up on return" with no local hint of which copy was the offender.

### Suggested shape

  Keep the destructor as a safety net (so an early return/throw doesn't leak), but never do throwing work in it, and require callers to submit() on the success path:
  ```cpp
  ~V2VBatch() {
    if (impl_ && !impl_->pending.empty()) {
      // Reaching the destructor with pending entries on a normal path is a bug.
      // Don't throw — we may be unwinding, and terminate() here would mask the
      // real error. Log loudly so it's caught in dev.
      if (std::uncaught_exceptions() == 0) {
        RBLN_LOG_WARN(
            "V2VBatch destroyed with {} pending entries — missing submit()",
            impl_->pending.size());
      }
      // Intentionally do NOT call submit() here: no throwing work in dtor.
    }
  }
  ```
  Call sites become:

  ```cpp
  c10::rbln::V2VBatch batch;
  for (...) strided_v2v_copy(dst, src, batch);
  batch.submit();   // copies are observable as complete after this line
  return out;
  ```

test/python/test_cat_index_select_v2v.py
rebel-chanheo: Thanks — all five addressed (markers, location, structure, fixture-based env, dedup).

aten/src/ATen/native/rbln/RBLNStridedV2V.cpp
rebel-chanheo: Agreed on all four. so i modified to explicit submit() on success path, dtor is warn-only and never issues backend calls.

test/rbln/test_cat_index_select_v2v.py
```diff
@@ -0,0 +1,550 @@
+# Owner(s): ["module: PrivateUse1"]
+
+"""Tests for the C++ v2v-based cat / index_select kernels.
```
rebel-jinhwan, line 3: Here, do we have regression tests that prevent v2h/h2v fallbacks and catch-up rsd issues?

rebel-chanheo: We can't honestly guard this from `test/rbln/`. The runtime's internal h2v/v2h happens below the PyTorch dispatcher, so it's invisible to op-level counters or `TorchDispatchMode`.
I understand that tests for this are already properly implemented in the rebel-compiler's unit tests, so I believe your concerns can be sufficiently addressed.

rebel-seungchul (approved):
LGTM

rebel-clee (approved):
LGTM

PR #50, opened 2026-05-13
fix(test): add to empty_cache before run unit-test

rebel-chanheo:
## Summary of Changes

  Adds an autouse pytest fixture (`reset_caching_allocator` in `test/conftest.py`) that calls
  `torch_rbln.memory.empty_cache(...)` for every visible RBLN device before each test, mirroring the existing
  `reset_dynamo` fixture. Without this, freed small blocks from prior tests accumulate in the caching allocator
  and fragment the pool, causing later tests (notably the model suite on `dev`) to fail when allocating large
  contiguous blocks.

  Provides a per-test opt-out via the `no_caching_allocator_reset` marker (registered in `pyproject.toml`) for
  any test that needs the allocator state preserved across the autouse boundary.

  ---

  ## Related Issues / Tickets

  * Resolves #

  ---

  ## Type of Change

  - [x] `fix` — Bug fix

  ---

  ## Affected Modules

  - [x] Build/Tools

  ---

  ## How to Test

  1. Run any subset of the suite under DEBUG logging to see the reset fire once per test, for every visible
  device:
     ```bash
     TORCH_RBLN_LOG_LEVEL=DEBUG pytest test/rbln/test_rbln_apis.py -q
     ```
     Expected: `Releasing RBLN caching allocator blocks on rbln:{0..N-1}` is logged before each test.

  2. Re-run the previously-failing model suite on `dev`:
     ```bash
     python test/run_tests.py --suite=models
     ```
     Expected: previously-failing tests pass (no fragmentation-induced large-block OOM).

  3. Opt-out path: any test decorated with `@pytest.mark.no_caching_allocator_reset` skips the reset, with a
  `Skipping RBLN caching allocator reset` debug log.

  ---

  ## Checklist

  * [x] Linting passes
  * [ ] All CI tests pass

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -370,5 +370,6 @@ markers = [
     "test_set_experimental: [test set] marks tests for experimental features (deselect with `-m \"not test_set_experimental\"`)",
     "test_set_perf: [test set] marks performance tests (select with `-m test_set_perf`)",
     "no_dynamo_reset: skips autouse TorchDynamo reset fixture",
+    "no_caching_allocator_reset: skips autouse RBLN caching-allocator reset fixture",
     "single_worker: marks tests that must run on a single worker (select with `-m single_worker`)",
 ]
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -3,8 +3,8 @@ import os
 import pytest
 import torch
 
+import torch_rbln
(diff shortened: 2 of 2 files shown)


rebel-clee (approved):
LGTM

PR #52, opened 2026-05-17
perf(eager): view-on-device align-fallback + BFS geometric-key cache

rebel-chanheo:
## Summary

Eager-mode perf optimizations on top of #46 (the foundation + view-on-device + fp16 fixes PR). Three commits that recover the LLaMA-1B rotary regression introduced by view-on-device, plus the BFS shape-key cache that brings prefill below the pre-regression sweet spot, plus a fp16 tolerance test for non-zero offset views.

This PR replaces the perf-only tail of the original PR #47 (which was auto-marked merged when `chan/perf-foundation` was force-pushed to include the original #47 head SHA in its history).

## What changes

**`perf(view-on-device): route 64-elem-unaligned single-tensor ops through cpu_fallback`** (`b134ac3`)

View-on-device emits `aten::expand/permute/narrow` on-device. For shapes whose last-dim element count is not a multiple of 64, rebel-compiler wraps the device fn with `contrib_aligned_pad` (pre-pad to 64) + `contrib_aligned_pad` (post-trim) + `contrib_dummy_cast` (f16↔dlf16) host-only ops with no RTOSA device-lowering. Each call pays the host-side `ApplyTensorDeviceTransform` memcpy + a full base-buffer D2H + H2D round-trip (~1 MB / op for LLaMA-1B rotary).

Routes single-tensor ops with `last_dim % 64 != 0` through `cpu_fallback_path` instead — one D2H of the source tensor + a host CPU op + one H2D of the result. **Recovers rotary_rotate_half decode by ~98%** without disabling view-on-device for the workloads where it pays off. Tensor-only (NOT list/tuple) — for list-input ops (cat, stack), the view-aware device path packs multiple unaligned tensors into a single fused IR that handles them cheaper than per-tensor host fallback (TensorList recursion would add +12% prefill via extra host-cat overhead).

**`perf(view-on-device): geometric-key cache for view recipe BFS`** (`8b33e37`)

Zero-base profile on the post-align-fallback HEAD identified `_bfs_search_recipe` as the dominant prefill cost: 48 BFS calls at ~93 ms cumulative each = **4.5 s of 12.9 s prefill (33%)**. The detector's output is a function of metadata alone (`data_ptr` / `Tensor._version` do not affect the search), so a geometric-only cache key `(base.size, base.stride, t.size, t.stride, t.storage_offset)` is sound.

A prior `G1` prototype keyed by `(data_ptr, _version, size, stride, offset)` never hit across layers (data_ptr differs per storage); the geometric key collapses LLaMA-1B's ~3 unique view patterns to 3 cache entries that 16 layers all hit.

Measured on rebel-compiler 045b3ed, LLaMA-1B prompt=32 / decode=8:
- Prefill wall:   12.92 s → **8.91 s** (**-31%**)
- Decode wall:    360.8 → 355.0 ms (-1.6%, within noise)
- BFS calls:      48 → 3 (-94%)
- Cache hit rate: **93.8%** (45 hits / 3 misses / 3 unique keys)

**`test(view-on-device): custom_float16-aware fp16 tolerance for non-zero offset tests`** (`0fa6593`)

Covers the non-zero `storage_offset` view path (the `narrow`-style recipes). Uses the `custom_float16_safe_range` mask from #46 to bound fp16 boundary mismatches at the edge of the `custom_float16` representable range — element-wise tolerance alone can't catch flush-to-zero / saturate-to-fp16-max effects on a sparse subset of elements.

## Risk + revert plan

Each commit is independent. The align-fallback (b134ac3) and BFS cache (8b33e37) are the higher-blast-radius commits; if either causes regressions on workloads outside the LLaMA-1B rotary path, revert individually.

### DIFF
--- a/test/rbln/test_non_zero_storage_offset.py
+++ b/test/rbln/test_non_zero_storage_offset.py
@@ -26,27 +26,6 @@ class BaseTestNonZeroStorageOffset(TestCase):
     # Flag to enable contiguous() call for non-zero storage offset tensors that are contiguous
     ALLOW_CONTIGUOUS_FOR_NONZERO_OFFSET = True
 
-    def assertEqual(self, x, y, *args, **kwargs):
-        # fp16 view-on-device adds a cast(f16↔cf16) round-trip on every view-aware
-        # op (`cast → strided_slice → cast`); custom_float16 has fewer mantissa bits
-        # than fp16, so each round-trip introduces ~1 ULP drift per element.
-        # Multi-stage view chains (chunk + transpose, slice + permute + reshape
-        # etc.) accumulate that drift across stages and trip the default-tight
-        # assertEqual tolerance even though the device computation is correct.
-        # Apply a custom_float16-aware tolerance for fp16 tensor compares — same scale
-        # the test_ops.py 30584d2 tolerance block uses for elementwise fp16 ops.
-        if (
-            "rtol" not in kwargs
-            and "atol" not in kwargs
-            and isinstance(x, torch.Tensor)
-            and isinstance(y, torch.Tensor)
-            and x.dtype is torch.float16
-            and y.dtype is torch.float16
-        ):
-            kwargs.setdefault("rtol", 0.005)
-            kwargs.setdefault("atol", 0.05)
-        return super().assertEqual(x, y, *args, **kwargs)
-
     def _check_storage_offset(self, tensor, expected_offset_min=0, msg=""):
         """Helper to check storage offset."""
         actual_offset = tensor.storage_offset()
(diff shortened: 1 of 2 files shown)


rebel-chanheo:
Closing — superseded by absorbing the 3 perf commits back into PR #46.

The split was wrong: the 64-elem align-fallback (`6d2d25d`) is the perf-recovery for view-on-device's rotary regression (+183% decode on LLaMA-1B without it). It belongs with view-on-device in the same PR. The BFS geometric-key cache (`9621167`) and the non-zero offset fp16 tolerance test (`d57d6dc`) are part of the same view-on-device perf story.

PR #46 head is now `dfa5799` (= original PR #47 head, contains all 3 of these commits as their original SHAs).

PR #54, opened 2026-05-20
feat(c10d): auto-discover RBLN_RDMA_IP in ProcessGroupRBLN ctor

rebel-jongwonlee:
## Summary of Changes

  Move RoCE v2 RDMA IP auto-discovery from the Python-side test helper
  (`torch_rbln._internal.rdma_env`) into the `ProcessGroupRBLN`
  constructor so production callers of `dist.init_process_group("rbln")`
  get the same `RBLN_RDMA_IP` behavior as the autoport tests without
  needing to call a Python helper explicitly.

  - Probes `/sys/class/infiniband` for a RoCE v2 capable device whose
    port is `ACTIVE` and whose netdev has an IPv4 matching the GID table,
    `PrepareContextAndExportMem()` (the point at which librbln-ccl reads
    the env).
  - Gated on `RCCL_PORT_GEN`: when autoport is enabled and the IP cannot
    be determined, fails fast with an actionable `RBLN_CHECK` instead of
    letting init blow up deep inside the runtime. When `RCCL_PORT_GEN`
    is unset, an unresolved `RBLN_RDMA_IP` is logged but non-fatal.
  - Escape hatch: `RBLN_DISABLE_AUTO_RDMA_IP=1` skips the probe entirely.

  ### Why this lives in `ProcessGroupRBLN` (not where you'd expect)

  The "right" architectural home for this is `librbln-ccl` itself
  (mirroring NCCL's `ncclIbGetGidIndex` / `ncclUpdateGidIndex` in
  `src/transport/net_ib.cc` — PyTorch's `ProcessGroupNCCL` does no
  network discovery and defers entirely to libnccl). Until the runtime
  gains that capability, the next-best layer is C++ at the PG boundary,
  because Python helpers only run when explicitly called.

  ### Designed for easy removal

  Everything new is isolated to a single removal unit so this can be
  deleted in one commit when librbln-ccl gains internal RoCE discovery:

  - New: `RdmaIpAutoDiscovery.{hpp,cpp}` (single removal target)
  - Modified: 3 lines in `CMakeLists.txt`, 8 lines in `ProcessGroupRBLN.cpp`
    (one include + one ctor call, both marked `TEMPORARY`)
  - The header documents the three-step removal procedure verbatim.
  - Verification command: `grep -rn "MaybeAutoDiscoverRbnRdmaIp\|RdmaIpAutoDiscovery\|RBLN_DISABLE_AUTO_RDMA_IP"`
    must return zero matches once removed.
  - `ProcessGroupRBLN.hpp` is **not** touched — no public API surface
    change, no member field added, no ABI risk.

  `torch_rbln._internal.rdma_env` and
  `test.utils.configure_rbln_network_for_autoport_tests` are
  intentionally left in place this round as defensive redundancy until
  the C++ path is exercised in CI.

  ---

  ## Related Issues / Tickets

  * Resolves (https://rbln.atlassian.net/browse/FINE-664)
  * Related to <librbln-ccl ticket TBD>

  ---

  ## Type of Change

  ---

  ## Affected Modules

  - [x] Backend
  - [x] Build/Tools

  ---

  ## How to Test

  1. Build: `./tools/dev-setup.sh external --clean`
  2. Autoport regression (existing tests):
     python -m test.distributed.test_process_group_autoport
     python -m test.distributed.test_tp_pp_autoport
  Both should pass — the Python helper still runs first, so the C++
  ctor sees `RBLN_RDMA_IP` already set and takes the no-op branch.
  3. C++ path alone (temporarily comment out
  `configure_rbln_network_for_autoport_tests()` in
  `test/utils.py`): re-run the same two tests; they should still pass
  on logs showing `[rbln_rdma_probe] RBLN_RDMA_IP=<ipv4> (auto-discovered)`
  coming from the C++ code path.
  4. Misconfig path: set `RCCL_PORT_GEN=1` on a host with no RDMA
  device (or `unshare --net` to drop `/sys/class/infiniband`) and
  construct a `ProcessGroupRBLN` — must raise `RBLN_CHECK` with the
  "RCCL_PORT_GEN is set but RBLN_RDMA_IP could not be determined"
  message rather than failing deep inside the runtime.
  5. Escape hatch: `RBLN_DISABLE_AUTO_RDMA_IP=1` + valid pre-set
  `RBLN_RDMA_IP` should pass; `RBLN_DISABLE_AUTO_RDMA_IP=1` +
  `RCCL_PORT_GEN=1` without `RBLN_RDMA_IP` should raise `RBLN_CHECK`.

  ---

  ## Screenshots / Logs

  Expected CI log line on success (matches the existing Python probe
  prefix so grep patterns keep working):

  [rbln_rdma_probe] RBLN_RDMA_IP=10.10.18.68 (auto-discovered)

  ---

  ## Checklist

  * [ ] Linting passes
  * [ ] All CI tests pass
  * [x] Relevant documentation has been updated (header comment documents removal procedure)

  ---

  ## Notes

  - **Temporary by design.** This is intentionally a workaround. The
    header carries explicit removal instructions; please do not extend
    this code or add new dependents — every additional caller raises the
    cost of the eventual removal.
  - **Namespace choice**: The helper lives in `torch_rbln::detail`
    (matching `ProcessGroupRBLNModule.hpp`), **not** `c10d::rbln::detail`.
    The latter shadows top-level `::rbln` (rebel-compiler runtime)
    inside `namespace c10d {}` and breaks every `rbln::MemoryInfo` /
    `rbln::DataType` reference. The header documents this constraint so
    a future refactor does not re-introduce the nesting.
  - **`std::once_flag`**: the probe runs once per process. If
    `RBLN_CHECK` throws, the flag is not consumed and a retry after env
    fix-up will work as expected.
  - Two commits intentionally separate (`feat` + `fix`) to leave the
    namespace-shadow learning visible in history; squash on merge if
    preferred.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,14 +12,13 @@ if(${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_BINARY_DIR})
 endif()
 
 # Set C++ and C standard
-set(CMAKE_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_C_STANDARD 11)
+set(CMAKE_C_STANDARD_REQUIRED ON)
 
 # Generate compile_commands.json
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
-
-set(CMAKE_LINK_WHAT_YOU_USE TRUE)
 
 # Set not to use link path for RPATH
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH OFF)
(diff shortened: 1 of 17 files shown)


rebel-seungchul:
@rebel-jongwonlee 
Could you add simple testcase for this feature?

rebel-seungchul (approved):
LGTM

PR #62, opened 2026-05-27
fix: harden strided-copy-based op kernels

rebel-clee:
## Summary of Changes

Hardens the three strided-copy-based op kernels (`cat`, `index_select`, `index_copy`) with an op-level CPU fallback and upstream-parity overlap checks. The fallback is enabled by default, gated on a new `strided_copy_error` category in `TORCH_RBLN_DISABLE_FALLBACK`.

The kernels have two robustness issues. First, the backend can fail on a large batched strided copy, breaking downstream model execution; the fallback keeps models running until the upstream backend is fixed. Second, the kernels lack the overlap checks that PyTorch's out variants enforce, so aliased inputs can produce silently incorrect results.

The fallback is narrowly scoped: only the batched-copy failure routes to CPU recomputation, while validation and other backend errors propagate so caller-side bugs still surface. The overlap checks exclude `index_copy`'s `self ↔ out` case, its legitimate in-place pattern.

---

## Type of Change

- [x] `fix` — Bug fix

---

## Affected Modules

- [x] Ops/Kernels
- [x] Docs

---

## Checklist

---

### DIFF
--- a/aten/src/ATen/native/rbln/RBLNStridedV2V.cpp
+++ b/aten/src/ATen/native/rbln/RBLNStridedV2V.cpp
@@ -2,10 +2,13 @@
 
 #include <ATen/native/rbln/RBLNStrideUtils.h>
 #include <ATen/native/rbln/RBLNTensorUtils.h>
+#include <c10/rbln/RBLNFallbackConfig.h>
 #include <c10/rbln/RBLNLogging.h>
 #include <c10/util/Exception.h>
 
 #include <cstdint>
+#include <functional>
+#include <string_view>
 #include <vector>
 
 namespace at::native::rbln {
@@ -108,7 +111,28 @@ void strided_v2v_copy(const at::Tensor& dst, const at::Tensor& src, c10::rbln::V
 void strided_v2v_copy(const at::Tensor& dst, const at::Tensor& src) {
   c10::rbln::V2VBatch batch;
   strided_v2v_copy(dst, src, batch);
-  batch.submit();
+  submit_or_fallback(batch, "strided_v2v_copy", [&] { dst.copy_(src.cpu()); });
+}
+
+void submit_or_fallback(c10::rbln::V2VBatch& batch, const char* op_name, std::function<void()> cpu_fallback) {
+  try {
+    batch.submit();
+  } catch (const c10::Error& e) {
+    const std::string_view error_message = e.what();
+    // TODO: Replace substring match with a typed exception when the wrapper API allows.
+    if (error_message.find("rbln_memcpy_v2v_multi failed") == std::string_view::npos) {
+      throw; // validation error — propagate
+    }
+    if (c10::rbln::is_fallback_disabled("strided_copy_error")) {
+      throw;
+    }
+    RBLN_LOG_WARN(
+        "{}: batched strided copy failed — falling back to CPU op. "
+        "Underlying error: {}",
+        op_name,
+        error_message);
+    cpu_fallback();
+  }
 }
 
 } // namespace at::native::rbln
--- a/aten/src/ATen/native/rbln/RBLNTensorShape.cpp
+++ b/aten/src/ATen/native/rbln/RBLNTensorShape.cpp
@@ -4,8 +4,10 @@
 #include <ATen/core/Tensor.h>
 #include <ATen/native/Resize.h>
 #include <ATen/native/rbln/RBLNStridedV2V.h>
+#include <ATen/ops/cat.h>
 #include <ATen/ops/empty.h>
 #include <c10/core/ScalarType.h>
+#include <c10/rbln/RBLNFallbackConfig.h>
 #include <c10/rbln/RBLNLogging.h>
 #include <c10/rbln/RBLNV2VBatch.h>
 #include <c10/util/Exception.h>
@@ -126,6 +128,12 @@ at::Tensor& cat_out_rbln(const at::ITensorListRef& tensors, int64_t dim, at::Ten
       c10::str(out.scalar_type()),
       c10::str(common_dtype));
 
+  // Mirror upstream PyTorch's `aten::cat.out` overlap checks.
+  at::assert_no_internal_overlap(out);
+  for (const auto& t : inputs) {
+    at::assert_no_overlap(out, t);
+  }
+
   // Resize via the upstream helper so a wrong-shape non-empty `out` triggers
   // the canonical UserWarning ("An output with one or more elements was
   // resized...").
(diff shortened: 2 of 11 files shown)


test/rbln/test_cat_index_select_v2v.py
```diff
@@ -306,6 +315,30 @@ def test_cat_mixed_dtype_promotes_to_common(self):
...
+        out = torch.empty(4, 6, dtype=dtype, device=DEVICE).t()
+        self.assertFalse(out.is_contiguous())
+        torch.cat([to_dev(a_cpu), to_dev(b_cpu)], dim=0, out=out)
+        _check(out, expected)
+
+    @pytest.mark.usefixtures("enable_eager_malloc")
+    def test_large_strided_cat(self):
+        """``aten::cat`` over a large batched strided copy."""
+        # Baseline allocations exercise the bulk strided copy path.
+        baseline_allocs = [torch.empty(102400, 2560, dtype=torch.float16, device=DEVICE) for _ in range(2)]
```
rebel-chanheo, line 333: How about adding a test case to test the environment variable strided_copy_error?

aten/src/ATen/native/rbln/RBLNStridedV2V.cpp
```diff
@@ -108,7 +109,18 @@ void strided_v2v_copy(const at::Tensor& dst, const at::Tensor& src, c10::rbln::V
 void strided_v2v_copy(const at::Tensor& dst, const at::Tensor& src) {
   c10::rbln::V2VBatch batch;
   strided_v2v_copy(dst, src, batch);
-  batch.submit();
+  try {
+    batch.submit();
+  } catch (const std::exception& e) {
```
rebel-chanheo: Exception scope is too broad here.

catch (const std::exception&) swallows everything derived from std::exception not just the "batched dispatch too large" failure this is meant to handle. With the fallback default-on, all of these get silently rerouted to CPU.

Two problems:
- It undercuts the project's stance (TORCH_RBLN_DISABLE_FALLBACK=compile_error) that dispatch/compile failures should be loud, not silently masked.
- On a hot path like cat/index_select/index_copy, a silent CPU roundtrip is a hidden perf cliff — only a WARN log, throughput just drops.

Suggestion: narrow the catch to the specific error the backend raises for this case (type / error code / message match) and rethrow the rest.

c10/rbln/RBLNV2VBatch.cpp
```diff
@@ -137,19 +137,29 @@ void V2VBatch::submit() {
...
+    } else {
+      // Heterogeneous — per-entry memcpy_v2v handles host-bounce internally.
+      RBLN_LOG_DEBUG("V2VBatch::submit draining {} entries (per-entry fallback)", impl_->pending.size());
+      for (const auto& e : impl_->pending) {
+        memcpy_v2v(e.dst, e.src, e.nbytes);
+      }
     }
+  } catch (...) {
+    reset_state();
+    throw;
```
rebel-chanheo: also, please check here

aten/src/ATen/native/rbln/RBLNTensorShape.cpp
```diff
@@ -175,7 +176,24 @@ at::Tensor& cat_out_rbln(const at::ITensorListRef& tensors, int64_t dim, at::Ten
     strided_v2v_copy(dst_view, t, batch);
     axis_offset += extent;
   }
```
rebel-seungchul, line 178: **Optional cleanup — duplicate submit/fallback block**

The `try { batch.submit(); } catch (...)` block here is repeated almost verbatim in `index_select_out_rbln` and `index_copy_out_rbln` (and a simpler variant in the 1-arg `strided_v2v_copy` overload). The only part that differs per call site is the CPU fallback body — the `is_fallback_disabled("strided_copy_error")` gate and the `RBLN_LOG_WARN` message are identical.

Consider factoring the common shell into a small helper in `RBLNStridedV2V.h` (all three TUs already include it), e.g.:

```cpp
template <typename Fn>
void submit_or_fallback(c10::rbln::V2VBatch& batch, const char* op_name, Fn&& cpu_fallback) {
  try {
    batch.submit();
  } catch (const std::exception& e) {
    if (c10::rbln::is_fallback_disabled("strided_copy_error")) {
      throw;
    }
    RBLN_LOG_WARN("{}: batched strided copy failed — falling back to CPU op. "
                  "Underlying error: {}", op_name, e.what());
    cpu_fallback();
  }
}
```

so each call site keeps only the op-specific logic:

```cpp
submit_or_fallback(batch, "cat_out_rbln", [&] {
  std::vector<at::Tensor> cpu_tensors;
  cpu_tensors.reserve(inputs.size());
  for (const at::Tensor& t : inputs) cpu_tensors.push_back(t.cpu());
  out.copy_(at::cat(cpu_tensors, axis));
});
```

This keeps the fallback policy (gate category + log wording) in one place, which helps if we later want to add a metric/counter on fallback. Non-blocking — feel free to skip if you'd rather keep the blocks inline.

(Note: the per-op overlap checks above are intentionally *not* worth factoring — they mirror upstream PyTorch's per-op `assert_no_overlap` argument sets 1:1, so spelling them out per kernel reads better.)

test/rbln/test_cat_index_select_v2v.py
rebel-clee: I omitted the test on purpose. Whether the intent is to verify the environment variable's routing on the happy path or to collect the cases that fail under it as expected failures, the same problem applies. Setting `TORCH_RBLN_DISABLE_FALLBACK=strided_copy_error` today triggers the rebel-compiler bug as an unhandled exception — exactly the diagnostic signal the environment variable produces. But a CI test that fails (or whose expected failure is the bug itself) goes stale the moment the fix lands.

When that fix lands, I plan to set `TORCH_RBLN_DISABLE_FALLBACK=strided_copy_error` as a default environment variable for the suite — same pattern as `TORCH_RBLN_DISABLE_FALLBACK=compile_error` — so the fallback path never engages silently in CI.

aten/src/ATen/native/rbln/RBLNStridedV2V.cpp
rebel-clee: Thanks — agreed the broad catch swallows everything `RBLN_CHECK` raises, including caller-side validation errors that should propagate (nullptr inputs, dtype mismatches, etc.). I'd ideally narrow by type via a typed exception, but that touches the wrapper API beyond this PR's scope.

As a near-term workaround I've added a stable "rbln_memcpy_v2v_multi failed" error message to the wrapper's `RBLN_CHECK`, and the fallback now matches on that substring — validation errors no longer hit the CPU path.

c10/rbln/RBLNV2VBatch.cpp
rebel-clee: Thanks for catching this. I've refactored `V2VBatch::submit()` to use an RAII guard for batch state reset — the guard's destructor now handles cleanup on both normal and exception paths, replacing the broad `catch (...)` and its catch-reset-rethrow wrapping.

aten/src/ATen/native/rbln/RBLNTensorShape.cpp
rebel-clee: Thanks — agreed the duplication is real. I've pulled `submit_or_fallback` into `RBLNStridedV2V.h` and `RBLNStridedV2V.cpp` along the lines you sketched. All three op kernels and the 1-arg overload now route through it.

rebel-seungchul (approved):
LGTM

rebel-chanheo (approved):
LGTM

PR #65, opened 2026-05-27
fix(compile): preserve outer chromium event stack across nested op-compile

rebel-jonghewk:
## Problem

With `VLLM_RBLN_USE_DEVICE_TENSOR=1`, model warm-up crashes during `torch.compile` guard building:

```
torch._dynamo.exc.InternalTorchDynamoError: RuntimeError: No toplevel event active.
Please only call this function within a metrics context/dynamo_timed.
```

(raised at `torch/_dynamo/guards.py` → `CompileEventLogger.increment_toplevel("guard_latency_us", ...)`)

## Root cause

Each RBLN ATen-op fallback (`register_ops.py`) is wrapped in `torch.compile` via `CompiledFunctionWrapper`. During an **outer** `torch.compile`'s `build_guards` phase, guard evaluation dispatches such an op, which **re-enters dynamo** (a nested `_compile`).

dynamo wraps every `_compile` in `chromium_event_timed("dynamo", reset_event_log_on_exit=True)`; on exit it calls `ChromiumEventLogger.reset()` → `stack.clear()` — the only stack-reset site in torch. Because that stack is thread-local and shared, the **nested** op-compile's exit wipes the **outer** compile's events too. The outer `build_guards` then calls `increment_toplevel("guard_latency_us")` on an empty stack → `No toplevel event active`.

Reproduces on torch 2.10 and is unchanged on torch `main` (guards.py / utils.py / convert_frame.py identical), so a torch upgrade does not fix it.

## Fix

`CompiledFunctionWrapper.__call__` snapshots the thread-local chromium event stack before running the (possibly nested) op-compile and restores it afterwards iff a nested compile wiped it. The outer compile's events survive and `guard_latency_us` is logged to the correct event. No-op when not inside an outer compile (≈ zero cost at inference).

## Testing

- **Unit (CPU):** without the guard the outer stack goes `['dynamo','build_guards'] → []` (crash); with it the stack is restored and `outermost == 'dynamo'`.
- **E2E:** MiniMax-M2, `VLLM_RBLN_USE_DEVICE_TENSOR=1`, DP=4 — warm-up passes `build_guards` and all 4 DP ranks complete generation (previously crashed at warm-up).

### DIFF
--- a/torch_rbln/_internal/torch_compile_patch_helpers.py
+++ b/torch_rbln/_internal/torch_compile_patch_helpers.py
@@ -35,38 +35,48 @@ def _exit_rbln_compile_op() -> None:
     _rbln_compile_op_depth.depth = max(0, d - 1)
 
 
-def _snapshot_chromium_event_state():
-    """Snapshot the thread-local chromium event stack so a nested op-compile's
-    ``reset_event_log_on_exit`` can't wipe an *outer* dynamo compile's events
-    (which crashes it in ``build_guards`` with "No toplevel event active").
+def _isolate_chromium_event_state():
+    """Isolate the thread-local chromium event state across a nested op-compile.
 
-    Each RBLN ATen-op fallback is itself ``torch.compile``-d, so dispatching one
-    during an outer compile re-enters dynamo. Returns a zero-arg restorer, or
-    ``None`` if there is nothing to protect.
+    Each RBLN ATen-op fallback is itself ``torch.compile``-d, so dispatching one during an
+    *outer* compile's ``build_guards`` re-enters dynamo. dynamo wraps every ``_compile`` in
+    ``chromium_event_timed(reset_event_log_on_exit=True)``; on exit it calls
+    ``ChromiumEventLogger.reset()``, which ``.clear()``s the *shared*, thread-local event
+    stack -- wiping the outer compile's events so ``build_guards`` crashes with
+    "No toplevel event active" when it logs ``guard_latency_us``.
+
+    Instead of repairing the damage afterwards, redirect the logger's thread-local
+    containers to fresh throwaways for the duration of the (possibly nested) op-compile and
+    restore the originals unconditionally. The nested compile then runs in -- and resets --
+    its own isolated containers, leaving the outer compile's state byte-identical. No-op
+    when not inside an outer compile (~zero cost at steady-state inference).
+
+    Returns a zero-arg restorer, or ``None`` if there is nothing to protect.
     """
     try:
         from torch._dynamo.utils import get_chromium_event_logger
 
         log = get_chromium_event_logger()
-        stack = log.get_stack()
-        if not stack:
+        tls = log.tls
+        if not log.get_stack():
             return None  # not inside an outer compile
-        saved_stack = list(stack)
-        saved_substack = list(log.get_pt2_compile_substack())
-        saved_event_data = dict(log.get_event_data())
+        # Hold the outer compile's live containers aside, untouched by the nested compile.
+        saved_stack = log.get_stack()
+        saved_substack = log.get_pt2_compile_substack()
+        saved_event_data = log.get_event_data()
     except Exception:
         return None
 
+    # The nested op-compile's reset_event_log_on_exit now clears these throwaways instead.
+    tls.stack = []
+    tls.pt2_compile_substack = []
+    tls.event_data = {}
+
     def _restore():
         try:
-            cur = log.get_stack()
-            if len(cur) >= len(saved_stack):
-                return  # not wiped
-            cur[:] = saved_stack
-            log.get_pt2_compile_substack()[:] = saved_substack
-            ed = log.get_event_data()
-            ed.clear()
-            ed.update(saved_event_data)
+            tls.stack = saved_stack
+            tls.pt2_compile_substack = saved_substack
+            tls.event_data = saved_event_data
         except Exception:
             pass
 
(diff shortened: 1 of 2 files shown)


torch_rbln/_internal/torch_compile_patch_helpers.py
```diff
@@ -35,6 +35,44 @@ def _exit_rbln_compile_op() -> None:
...
+                return  # not wiped
+            cur[:] = saved_stack
+            log.get_pt2_compile_substack()[:] = saved_substack
+            ed = log.get_event_data()
+            ed.clear()
+            ed.update(saved_event_data)
+        except Exception:
+            pass
+
+    return _restore
```
rebel-seungchul: **Suggestion: isolate the event state instead of repairing it afterwards.**

The current `_snapshot_chromium_event_state` lets the nested op-compile run on the *shared* containers and then conditionally repairs them with the heuristic `len(cur) >= len(saved_stack)` ("not wiped"). A more robust, deterministic approach is to redirect the logger's thread-local containers to fresh throwaways *before* the nested compile and restore the originals **unconditionally** afterwards.

`ChromiumEventLogger.reset()` mutates the lists in place via `.clear()`, and `get_stack()` reads `self.tls.stack` each time, so swapping `tls.stack` / `tls.pt2_compile_substack` / `tls.event_data` fully isolates the nested compile:

- No `len(...)` heuristic; restore is unconditional and the outer state is byte-identical (the originals are never touched).
- The nested op-compile's events never mix into the outer compile's chromium trace.
- Keeps NPU execution (no CPU fallback); still a no-op outside an outer compile.

```suggestion
def _isolate_chromium_event_state():
    """Isolate the thread-local chromium event state across a nested op-compile.

    Each RBLN ATen-op fallback is itself ``torch.compile``-d, so dispatching one during an
    *outer* compile's ``build_guards`` re-enters dynamo. dynamo wraps every ``_compile`` in
    ``chromium_event_timed(reset_event_log_on_exit=True)``; on exit it calls
    ``ChromiumEventLogger.reset()``, which ``.clear()``s the *shared*, thread-local event
    stack -- wiping the outer compile's events so ``build_guards`` crashes with
    "No toplevel event active" when it logs ``guard_latency_us``.

    Instead of repairing the damage afterwards, redirect the logger's thread-local
    containers to fresh throwaways for the duration of the (possibly nested) op-compile and
    restore the originals unconditionally. The nested compile then runs in -- and resets --
    its own isolated containers, leaving the outer compile's state byte-identical. No-op
    when not inside an outer compile (~zero cost at steady-state inference).

    Returns a zero-arg restorer, or ``None`` if there is nothing to protect.
    """
    try:
        from torch._dynamo.utils import get_chromium_event_logger

        log = get_chromium_event_logger()
        tls = log.tls
        if not log.get_stack():
            return None  # not inside an outer compile
        # Hold the outer compile's live containers aside, untouched by the nested compile.
        saved_stack = log.get_stack()
        saved_substack = log.get_pt2_compile_substack()
        saved_event_data = log.get_event_data()
    except Exception:
        return None

    # The nested op-compile's reset_event_log_on_exit now clears these throwaways instead.
    tls.stack = []
    tls.pt2_compile_substack = []
    tls.event_data = {}

    def _restore():
        try:
            tls.stack = saved_stack
            tls.pt2_compile_substack = saved_substack
            tls.event_data = saved_event_data
        except Exception:
            pass

    return _restore
```

torch_rbln/_internal/torch_compile_patch_helpers.py
```diff
@@ -310,9 +348,13 @@ def _handle_compile_exception(self, error, args, kwargs, *, is_recompile_limit,
     def __call__(self, *args, **kwargs):
         """Execute the compiled function with reentrancy guard, TP auto-determination and failover."""
         _enter_rbln_compile_op()
+        # Protect an outer compile's chromium event stack from this op-compile's reset.
+        restore_chromium = _snapshot_chromium_event_state()
```
rebel-seungchul: Rename the call site to match the isolation helper:

```suggestion
        # Isolate an outer compile's chromium event stack from this op-compile's reset.
        restore_chromium = _isolate_chromium_event_state()
```

rebel-seungchul (commented):
One suggestion on the *mechanism*: rather than snapshotting and then conditionally restoring the shared event containers, we can **isolate** them for the duration of the nested op-compile and restore unconditionally. This removes the `len(cur) >= len(saved_stack)` heuristic, guarantees the outer compile's state is byte-identical afterwards, and prevents the nested op-compile's events from leaking into the outer compile's chromium trace. Details inline.

I've validated the isolated version locally against a real nested `torch.compile` (outer metrics context + a `"dynamo"` chromium toplevel): the outer event survives and `CompileEventLogger.increment_toplevel("guard_latency_us")` no longer raises. A reproduction test (driving the real RBLN op fallback on `rbln` tensors during `build_guards`) is on hold and will follow in a later commit.

rebel-chanheo (approved):
~~LGTM~~

rebel-chanheo (changes requested):
this PR makes regression performance. we need to be talk about this in offline.

rebel-jonghewk:
@rebel-chanheo Thanks for flagging — pinned it down to the per-op-call fast-path. `CompiledFunctionWrapper.__call__` runs for every RBLN op dispatch (including steady-state inference, where no outer compile is in flight); the previous fast-path was paying for a function-local `from torch._dynamo.utils import ...` (sys.modules dict + attr lookup) and two `log.get_stack()` method calls (one of which lazily allocates `tls.stack` as a side effect).

`4411dd3` caches the chromium-logger getter at module load and reads `tls.stack` via `getattr(tls, "stack", None)`, so the fast-path is a single attribute lookup. Slow-path (during an outer compile) behavior is unchanged: throwaway swap + unconditional restore.

CPU microbench (1M calls, no outer compile in flight):

| | ns/call |
| --- | --- |
| before | 539 |
| after `4411dd3` | 128 |
| reduction | **−411 (−76%)** |

Happy to still discuss offline if there's more concern.

rebel-chanheo (approved):
LGTM

rebel-seungchul (approved):
LGTM

PR #76, opened 2026-06-04
perf(copy): route strided copy_ through on-device v2v engine

rebel-jonghewk:
## Description

Part of fixing slow device-tensor KV d2d copy (rebellions-sw/vllm-rbln-internal#94). Strided `aten::copy_` on RBLN tensors fell back to a host bounce (device→host→device of the full view span), so KV sub-block copies paid that round-trip on every layer. This routes them to the on-device strided v2v engine when the geometry favors it.

1. **Routing:** non-contiguous, same-shape/dtype, same-device rbln→rbln `copy_` now goes through `strided_v2v_copy` (→ `rbln_memcpy_v2v_multi`) instead of the host bounce. Contiguous copies keep the existing direct `memcpy_v2v` path.
2. **Gate:** routed only when `outer_count <= 1024` (computed in O(rank) from strides, no expansion). Above the cap the runtime per-dst v2v limit (`kMaxV2VMultiCopies`) would fan out to a host fallback anyway, so we bounce via host directly and skip the wasted descriptor expansion.
3. **Paired PR:** rebel_compiler raises `kMaxV2VMultiCopies` to match the 1024 gate — rebellions-sw/rebel_compiler#11250. Both must land together.

### DIFF
--- a/aten/src/ATen/native/rbln/RBLNCopy.cpp
+++ b/aten/src/ATen/native/rbln/RBLNCopy.cpp
@@ -1,6 +1,8 @@
 // TODO: The previous copy optimizations based on physical shape and physical dtype were removed during
 // v-memory integration. Revisit these optimizations in a future pass.
 #include <ATen/native/rbln/RBLNCopy.h>
+#include <ATen/native/rbln/RBLNStrideUtils.h>
+#include <ATen/native/rbln/RBLNStridedV2V.h>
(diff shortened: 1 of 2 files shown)


rebel-chanheo (approved):
LGTM

rebel-clee (approved):
LGTM

PR #55, opened 2026-05-20
perf(native): replace cpu_fallback with native impls for fill_/arange/_local_scalar_dense/clone

rebel-chanheo:
## Summary of Changes

Direct PrivateUse1 impls for four ATen ops that previously fell through to
`cpu_fallback_rbln`, eliminating the `redispatchBoxed(CPU)` + `TensorIterator`
overhead. Each op is hit on every vLLM decode step.

- `aten::fill_.Scalar` — host borrow + typed `std::fill_n` / `std::memset`; CPU fallback for non-contig / unsupported dtype.
- `aten::arange.start_out` — host range fill + `acquire_host_ptr_for_overwrite` + lazy h2v; CPU fallback for unsupported dtypes (Half/BFloat16/complex).
- `aten::_local_scalar_dense` — direct vmem read for `tensor.item()`.
- `aten::clone` — `memcpy_v2v` fast path for dense-storage source; `Preserve` memory format resolved per source layout (contig / strided / channels_last).

## Related Issues / Tickets

* Related to #

## Type of Change

- [x] `core` — RBLN backend op impls
- [x] `perf` — host-overhead reduction

## Affected Modules

- [x] Backend
- [x] Ops/Kernels
- [x] C++ extension (`torch_rbln/csrc`)

## How to Test

```bash
pytest test/rbln/test_native_fill_scalar.py \
       test/rbln/test_native_arange.py \
       test/rbln/test_native_local_scalar_dense.py \
       test/rbln/test_native_clone.py
python test/run_tests.py --suite=ops --test_mode=release
```

65 native-op cases (full dtype matrix + edge cases) + 1627 release-mode ops tests pass.

## Notes

Stacked on `chan/perf-foundation` (PR #46).

PR #78, opened 2026-06-05
docs(fallback): tighten verbose comments in cpu_fallback_rbln

rebel-chanheo:
## Summary

Tightens the longest comment blocks in `RBLNCPUFallback.cpp` so the logic stays in the foreground (follow-up to the review note on #46). Keeps the borrow-safety rationale and the dated/profiling anchors; cuts restatement and padding.

Blocks compressed: schema cache, Step 1-4 headers, pure-`out=` fill, direct output borrow, RAII release guard, view-op note.

## Scope

Comments only — **no code change**. 812 -> 758 lines (109 deletions / 55 insertions).

## Test

N/A (comments only). Diff verified to touch only `//` lines.

### DIFF
--- a/aten/src/ATen/native/rbln/RBLNCPUFallback.cpp
+++ b/aten/src/ATen/native/rbln/RBLNCPUFallback.cpp
@@ -31,19 +31,12 @@
 namespace at::native::rbln {
 
 namespace {
-// Per-op schema cache: mirrors the DispatchShim SchemaCache idea. Same op handle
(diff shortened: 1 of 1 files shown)


rebel-chanheo:
Folded into #55 (chan/perf-foundation-native-ops) as f63f303 — the comment cleanup now lives with the native-impl work rather than as a standalone PR off dev. Closing this one.

PR #82, opened 2026-06-08
refactor: centralize supported dtype policy

rebel-clee:
## Summary of Changes

Refactors the dtype dispatch surface so adding a new dtype becomes a small, mechanical change. Consolidates the catalog into a single C++ header (`c10/rbln/RBLNSupportedDtypes.h`) and binds it to Python via `torch_rbln._C`.

Only float16 is admitted today, so the catalog refactor is behavior-preserving. The C++ header is the source of truth — the dispatch predicate stays `constexpr` and the tuples are cached at import; the alternative — a Python source of truth — would force a runtime lookup on every dispatch.

Also bundles three drive-by fixes. Two are latent float16-pinning in paged attention and the `AllReduce` op — no-ops today. The third fixes an active SDPA-fallback bug in `_sdpa_attn_weights_fallback`: a bool `attn_mask` left masked positions at `0.0` instead of `-inf`.

---

## Type of Change

---

## Affected Modules

- [x] Backend
- [x] Device
- [x] Ops/Kernels
- [x] Docs

---

## Checklist

---

### DIFF
--- a/c10/rbln/CMakeLists.txt
+++ b/c10/rbln/CMakeLists.txt
@@ -19,6 +19,7 @@ set(headers
   RBLNHooksInterface.h
   RBLNLogging.h
   RBLNMacros.h
+  RBLNSupportedDtypes.h
   RBLNV2VBatch.h
 )
 add_library(${target}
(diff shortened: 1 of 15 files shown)


rebel-chanheo (approved):
LGTM

PR #83, opened 2026-06-08
perf(rbln): add CPU fast-path handlers for int sub/mul/clamp metadata ops

rebel-chanheo:
## Summary of Changes

The fp16-only RBLN device cannot run integer element-wise ops natively, so `aten::sub.out` / `aten::mul.out` / `aten::clamp.out` on int16/int32/int64 tensors fall back to CPU via `redispatchBoxed(CPU)`. For the tiny integer tensors that appear every decode step in the attention metadata builder (partition-size math and `logits_indices`), that path's boxing + `TensorIterator` setup dominates the cost.

This adds host micro-kernels for those three ops to the existing `CPUFastPathRegistry` (same pattern as the `rsqrt`/`pow`/`mean` fast paths). The handlers:

- compute in int64 and store in the output dtype, so mixed-dtype scalar promotion (e.g. `int32 * <int64 scalar>`) is handled correctly;
- support scalar / same-shape / last-dim broadcast;
- bail to the boxed fallback on any unsupported shape/dtype, so results are never wrong — only un-accelerated.

Files: new `aten/src/ATen/native/rbln/fast_paths/int_elementwise.cpp` + its entry in `aten/src/ATen/CMakeLists.txt`.

## Related Issues / Tickets

* Related to #

## Type of Change

## Affected Modules

- [x] Backend
- [x] Ops/Kernels

## How to Test

1. `pytest test/rbln/test_cpu_fast_paths.py` — added registration checks for `aten::sub.out` / `mul.out` / `clamp.out` plus numerical-correctness tests (same-shape, scalar / last-dim broadcast, int16/int32/int64, and the `alpha != 1` guard fall-through), mirroring the existing rsqrt/pow/mean coverage.
2. On Llama-3.2-1B decode, the per-step `cpu_fallback` wall time drops from ~117 to ~78 us/step (sub/mul/clamp no longer pay `redispatchBoxed`).

## Checklist

* [x] Linting passes (clang-format)
* [ ] All CI tests pass
* [x] The test method is described — unit tests added in `test/rbln/test_cpu_fast_paths.py`

## Notes

Pure addition — no existing op behavior changes. Unregistered ops and unsupported shapes/dtypes keep the current boxed fallback path.

### DIFF
--- a/aten/src/ATen/CMakeLists.txt
+++ b/aten/src/ATen/CMakeLists.txt
@@ -14,6 +14,7 @@ set(sources
   native/rbln/fast_paths/rsqrt_fp32.cpp
   native/rbln/fast_paths/pow_squared_fp32.cpp
   native/rbln/fast_paths/mean_lastdim_fp32.cpp
+  native/rbln/fast_paths/int_elementwise.cpp
 )
 set(headers
   native/rbln/RBLNCPUFallback.h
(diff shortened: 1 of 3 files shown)


rebel-clee (approved):
LGTM

PR #85, opened 2026-06-09
fix: green models suite on RBLN-CR03 (REBEL); graceful v2v fallback

rebel-chanheo:
## Summary

Make `test/models` pass on **RBLN-CR03 (REBEL)** and harden the on-device v2v copy
fallback. Verified on CR03: **`pytest test/models -m test_set_ci` → 21 passed,
6 skipped (TP>1), 0 failed** (27 min).

## Changes

**1. KV-validation `str()` works on torch 2.10 and 2.11** (`test_optimum_llm.py`)
`_validate_kv_layer_after_prefill` prints a CPU copy (`layer_tensor.to("cpu")[...]`).
`str()` of an RBLN-device tensor runs the formatter, which on torch 2.11 can route a
`.out` op (e.g. `aten::mul.out`) through cpu_fallback and raise `... invalid alias
information ... Tensor(a! -> a)`. Values are still checked by the `copy_`/`allclose`
just above.

**2. Skip TP>1 on the REBEL (CR) lineup** (`test_optimum_llm.py`, `test_vllm_llm.py`, `test/utils.py`)
New `is_rebel_device()` (RBLN-CR03/CR13/CR23 vs ATOM RBLN-CA*). A single REBEL device is
a quad chiplet (1→4) and TP>1 isn't supported there yet (vllm-rbln itself asserts
`VLLM_RBLN_TP_SIZE == 1` for `"cr"`). ATOM still runs TP2.

**3. REBEL-conditional vLLM `gpu_memory_utilization` = 0.1** (`test_vllm_llm.py`)
REBEL has 140 GB DRAM, so the ATOM-tuned `0.5` (=70 GB) over-allocates the KV pool to
~66 GiB / 2117 blocks for a 1B model — which triggers the v2v failure below and then OOM.
`0.1` (~10 GiB / 325 blocks) is ample for these single-sequence cases. (num_blocks scales
with available memory, **not** `max_model_len` — lowering max-seq-len does not help.)

**4. Graceful CPU fallback when v2v rejects a copy** (`RBLNFunctions.cpp`, `RBLNStridedV2V.cpp`)
`submit_or_fallback` only matched the batched `rbln_memcpy_v2v_multi failed`; when
`V2VBatch::submit` drains per-entry, the same-device `memcpy_v2v` failure was message-less
and escaped the gate as a hard crash. Tag it (`rbln_memcpy_v2v failed`) and match it so a
rejected v2v falls back to a host bounce (`dst.copy_(src.cpu())`). Inert when v2v succeeds
(e.g. `gpu_mem=0.1`).

## v2v failure — root cause (runtime follow-up)

At `gpu_mem=0.5` the eager naive-prefill KV write hit `rbln_memcpy_v2v → rc=1`
(RBLNRetCode_FAILURE). `rbln_get_memory_info` on the failing 128 B same-device copy:
- **DST** `0x8ffd24761`: `key_vaddr=0x8ffc24761` (**interior +1 MiB offset**),
  `user_dtype=Undefined`, empty shape → an untyped/raw pooled allocation (vLLM's flat KV buffer).
- **SRC** `0x1158207ead`: typed (`key_vaddr==vaddr`) but `user_dtype=Float16` /
  `physical_dtype=BFloat16`.

A standalone replica (typed `torch.zeros`, `key_vaddr==vaddr`) takes v2v fine even at
**64 GB vaddr depth / 66 GiB total / cross-chiplet** — so it is **not** size, vaddr-depth,
total-memory, or chiplet-crossing. The runtime v2v resolves typed base allocations but
**rejects interior offsets into large untyped pool allocations**. Runtime follow-up: v2v
should handle those (and/or the fp16/bf16 source mismatch); meanwhile change 3 avoids the
regime and change 4 degrades gracefully.

## Out of scope / env
- TP>1 on REBEL (chiplet) and the runtime v2v limitation above are tracked separately.
- Reproducing the green suite requires **rebel-compiler 0.10.5.dev162.prod**.

### DIFF
--- a/aten/src/ATen/native/rbln/RBLNStridedV2V.cpp
+++ b/aten/src/ATen/native/rbln/RBLNStridedV2V.cpp
@@ -120,7 +120,13 @@ void submit_or_fallback(c10::rbln::V2VBatch& batch, const char* op_name, std::fu
   } catch (const c10::Error& e) {
     const std::string_view error_message = e.what();
     // TODO: Replace substring match with a typed exception when the wrapper API allows.
-    if (error_message.find("rbln_memcpy_v2v_multi failed") == std::string_view::npos) {
+    // Route both v2v backend rejections to the CPU fallback: the batched path
+    // throws "rbln_memcpy_v2v_multi failed", and when it drains per-entry the
+    // same-device path throws "rbln_memcpy_v2v failed". (The runtime rejects v2v
+    // into interior offsets of large untyped pool allocations — e.g. vLLM's KV
+    // cache; the host-bounce CPU fallback writes those correctly.)
+    if (error_message.find("rbln_memcpy_v2v_multi failed") == std::string_view::npos &&
+        error_message.find("rbln_memcpy_v2v failed") == std::string_view::npos) {
       throw; // validation error — propagate
     }
     if (c10::rbln::is_fallback_disabled("strided_copy_error")) {
--- a/c10/rbln/RBLNFunctions.cpp
+++ b/c10/rbln/RBLNFunctions.cpp
@@ -306,7 +306,11 @@ void memcpy_v2v(void* rbln_dst_data, const void* rbln_src_data, size_t nbytes) {
     RBLN_LOG_DEBUG("Performing same-device copy");
 
     RBLN_LOG_DEBUG("Calling rbln_memcpy_v2v: src_vaddr={:#x}, dst_vaddr={:#x}, size={}", src_vaddr, dst_vaddr, size);
-    RBLN_CHECK(!::rbln::rbln_memcpy_v2v(src_vaddr, dst_vaddr, size));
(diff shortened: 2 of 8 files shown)


rebel-chanheo:
Update: standalone repro **found** after all (earlier comment outdated). Trigger isolated: a strided write from an `expand().contiguous()` copy of a compiled-op output into a pool **past the ~34 GiB chiplet boundary** fails (rc=1) once chiplet 0 has been filled with multiple strided-touched lazy pools (pool size irrelevant — 1 GiB ×34 or 4 GiB ×8; a single touched 33 GiB pool does NOT trigger). Added as `test/rbln/test_v2v_cross_chiplet.py` — it triggers the rejection deterministically and pins value-correctness via the CPU fallback; also serves as the runtime-team repro.

rebel-clee (approved):
LGTM

PR #86, opened 2026-06-10
feat(memory): support pinned host memory and torch.Event

rebel-chanheo:
## Summary of Changes

This PR adds pinned host memory and `torch.Event` support to the RBLN backend, completing the `pin_memory` + `non_blocking` copy path. Motivation: vllm-rbln currently disables `is_pin_memory_available()` and keeps its `transfer_event` record/synchronize lines commented out because RBLN lacked both pieces (rebellions-sw/vllm-rbln-internal#122); enabling them is a vllm-rbln follow-up PR.

**Problem 1 — pin_memory crashes.** Importing `torch_rbln` registers RBLN as the PrivateUse1 accelerator, which routes every CPU `pin_memory=True` allocation to the PrivateUse1 hooks' `getPinnedMemoryAllocator()`. RBLN did not implement it, so **creating a plain CPU tensor with `pin_memory=True` raised `NotImplementedError`**.

**Problem 2 — pageable non_blocking host-read race.** #13 dispatches `non_blocking=True` async regardless of the host memory type. Reading a pageable dst without `torch.rbln.synchronize()` races (no auto-drain on host read; rebellions-sw/rebel_compiler#11343 covers device-side hazards only). vllm-rbln has 27 such call sites today. Gating on pinned was deferred in the #13 review solely because no pinned allocator existed.

**Problem 3 — no event to fence an async copy.** Once pinned copies are truly async, the caller must fence before reading. The device-agnostic pattern is `pinned.copy_(t, non_blocking=True); event.record(); event.synchronize()` — but `torch.Event` raised "Backend doesn't support events" on RBLN. Whole-device `torch.rbln.synchronize()` existed, but events are what device-agnostic callers (vLLM, torch.accelerator) actually use.

**Solution.**

| Commit | Change |
|---|---|
| `feat(memory)` | `RBLNPinnedAllocator`: page-aligned + best-effort `mlock`; base→size registry backs `isPinnedPtr` (interior pointers / views work); hooks overrides |
| `feat(copy)` | Gate non_blocking on host memory type, matching PyTorch's documented CUDA semantics: pageable host → sync (safe to read without synchronize); pinned host → async (caller fences) |
| `feat(event)` | `torch.Event`: single in-order copy queue → an event records its device, every wait maps to `synchronize(device)`. Also `synchronizeDevice`, so `torch.accelerator.synchronize()` works |

```
non_blocking=True ──► host pinned? ──► yes ──► async dispatch (record event, sync before read)
                              └─────► no  ──► sync copy (read immediately)
```

Design notes:
- UMD has no DMA-mapped host alloc yet; stage 1 is `mlock` page locking. When the API lands only allocate/free change.
- DataPtr device is CPU — pinned tensors keep standard semantics.
- Events carry no timing (`elapsed_time` raises); `query()` drains then returns true, so it may block — review polling loops when adopting.
- Event waits drain the whole device queue (no per-event fences); correct by queue order, just over-synchronizes.
- Streams stay default-only; per-stream events and `wait_stream` are no-ops by queue order. Separate issue if multi-queue ever lands.

---

## Related Issues / Tickets

* Resolves rebellions-sw/vllm-rbln-internal#122
* Related to #13 (stacked PR — retarget to dev after #13 merges)

---

## Type of Change

---

## Affected Modules

- [x] Backend
- [x] Memory

---

1. Issue repro: `python -c "import torch_rbln, torch; t = torch.tensor([[1],[2],[3]], dtype=torch.long, pin_memory=True); print(t.is_pinned())"` → `True` (previously NotImplementedError)
2. `pytest test/rbln/test_pin_memory.py` → 14 passed (incl. pageable non_blocking read-without-sync correctness)
3. `pytest test/rbln/test_event.py` → 7 passed (incl. pinned D2H → record → synchronize → tolist)
4. `ctest --test-dir ./build -R Pinned` → 4 passed; `test_cpp_RBLNEventTest` → 3 passed
5. Regression: `pytest test/rbln/test_pin_memory.py test/rbln/test_tensor_copy.py test/rbln/test_rbln_apis.py` → 328 passed

Verified with rebel-compiler==0.10.5.dev187+g95f45d1f, ATOM, torch 2.11.0+cpu.

---

## Checklist

* [x] Linting passes (ruff, clang-format)
* [ ] All CI tests pass

---

## Notes

The gate downgrades today's pageable non_blocking copies (1.35–2.5x in #13) to sync until callers pin. This trades that speedup for correctness-by-construction; perf-sensitive paths opt back in via `.pin_memory()` + event sync.

Reviewer — Primary: (TBD) / Secondary: (code owners). Please assign a primary reviewer before merge.

### DIFF
--- a/aten/src/ATen/native/rbln/RBLNCopy.cpp
+++ b/aten/src/ATen/native/rbln/RBLNCopy.cpp
@@ -13,7 +13,6 @@
 #include <rebel/runtime/api/rbln_runtime_api.h>
 
 #include <cstddef>
-#include <cstdlib>
 
 namespace at::native::rbln {
 
@@ -206,19 +205,11 @@ void copy_impl_rbln_async(const at::Tensor& src, const at::Tensor& dst) {
     return;
   }
 
-  // CUDA semantics: async dispatch only when the host side is pinned. A
-  // pageable host tensor is read/written by plain CPU code that never passes
-  // through RBLN, so nothing can drain a still-pending transfer before the
-  // access; downgrading to a sync copy makes such code correct by
-  // construction. Pinned host memory opts in to true async (caller
-  // synchronizes before touching it). Set TORCH_RBLN_PAGEABLE_ASYNC=1 to
-  // restore unconditional async dispatch during migration.
-  static const bool pageable_async = []() {
-    const auto* env = std::getenv("TORCH_RBLN_PAGEABLE_ASYNC");
-    return env != nullptr && env[0] == '1';
-  }();
+  // CUDA semantics: async only when the host side is pinned. Host reads never
+  // pass through RBLN, so nothing can drain a pending transfer first — sync
+  // for pageable makes that safe; pinned opts in (caller synchronizes).
   const at::Tensor& host = src_device.is_cpu() ? src : dst;
-  if (!pageable_async && host.device().is_cpu() && !c10::rbln::is_pinned_ptr(host.data_ptr())) {
+  if (host.device().is_cpu() && !c10::rbln::is_pinned_ptr(host.data_ptr())) {
     RBLN_LOG_DEBUG("Pageable host tensor, downgrading non_blocking copy to sync");
     copy_impl_rbln(src, dst);
     return;
--- a/aten/src/ATen/native/rbln/RBLNCopy.cpp
+++ b/aten/src/ATen/native/rbln/RBLNCopy.cpp
@@ -9,9 +9,11 @@
 #include <c10/rbln/RBLNFallbackConfig.h>
 #include <c10/rbln/RBLNFunctions.h>
 #include <c10/rbln/RBLNLogging.h>
+#include <c10/rbln/RBLNPinnedAllocator.h>
 #include <rebel/runtime/api/rbln_runtime_api.h>
 
 #include <cstddef>
+#include <cstdlib>
 
 namespace at::native::rbln {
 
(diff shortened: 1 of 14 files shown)


rebel-chanheo:
@rebel-jonghewk, could I ask for your review this PR when you have a moment?

The primary motivation is rebellions-sw/vllm-rbln-internal#122: once `torch_rbln` is imported, RBLN becomes the `PrivateUse1` accelerator and every CPU `pin_memory=True` allocation routes to our hooks — with `getPinnedMemoryAllocator` unimplemented, even plain CPU tensor creation raises `NotImplementedError`. 
#86 adds the pinned-host allocator (`getPinnedMemoryAllocator` / `isPinnedPtr`) to make the standard pin_memory UX work end to end.

This PR stacks on #13 and also follows up on the review thread we left there. Now that a pinned allocator exists, the precondition behind deferring the pinned gating ("RBLN has no pinned-host allocator") no longer applies, so the gate is included as well:

- **Pinned host memory → true** async dispatch (caller synchronizes before reading, same contract as #13)
- **Pageable host memory → always-sync copy** — the async path #13 enabled for pageable now downgrades to sync

That means code like:

```
x = t.to("cpu", non_blocking=True)   # pageable dst → sync copy
use(x)                                # safe immediately, no synchronize() needed

p = t.cpu().pin_memory()
p.copy_(t, non_blocking=True)         # pinned dst → async
torch.rbln.synchronize()              # caller's responsibility
use(p)
```

becomes correct by construction for pageable, while pinned keeps the full async speedup. This matches CUDA exactly — CUDA forces sync for pageable non_blocking and only goes async on pinned hosts. CUDA-habit code (vllm-rbln currently has ~27 pageable non_blocking sites with no sync before host reads) works unmodified.

The trade-off is that today's pageable callers lose the #13 async speedup until they opt in via .pin_memory() — for vllm-rbln a single is_pin_memory_available() flip restores it (with sync re-enabled before host reads).

Would appreciate your thoughts on the semantic change — happy to discuss alternatives if you see issues.

rebel-jonghewk (approved):
lgtm

PR #93, opened 2026-06-16
fix: chunk AllGather to a CS-safe size to fix REBEL tail corruption

rebel-jongwonlee:
## Summary

On **REBEL (CR03)**, `dist.all_gather` corrupted the **tail** of large per-rank tensors — data past ~63 MiB came back stale/swapped. Deterministic, independent of rebel-compiler version. Tracked in **rebellions-sw/fsw-inference#324**.

## Root cause

It is **not** an alignment requirement per se — it's a **command-stream (CS) size limit** in the UMD AllGather:

- The whole AllGather CS is generated into a single fixed `RCCL_CS_CHUNK_SIZE` (**32 KiB**) chunk, and exactly that many bytes are shipped to the device (`allgather.c`: `*cs_size = RCCL_CS_CHUNK_SIZE; rccl_h2d_memcpy(..., RCCL_CS_CHUNK_SIZE)`; both ring and reverse-recursive-doubling paths).
- A transfer large enough that its per-2 MiB-page DMA descriptors overflow 32 KiB has its **tail descriptors dropped** → recv-buffer tail never DMA'd (left stale), or the op errors with `-ENOMEM`.
- CS size scales with the per-call **output** bytes. Empirically (world_size=2): **256 MiB output overflows, 128 MiB is safe**. A non-2 MiB-aligned tail adds one extra descriptor, which is why a borderline *aligned* size passes while the *unaligned* one fails.

## Fix

Lower `RCCL_ALLGATHER_MAX_OUTPUT_BYTES` **1 GiB → 64 MiB** so large transfers are split through the **existing chunked multi-call path**, keeping each call's CS well within the 32 KiB budget.

- No alignment change → **no small-tensor padding overhead**.
- A 2 MiB-alignment workaround alone would *not* help large **aligned** transfers, which still overflow the CS — chunking is the correct fix.
- 64 MiB leaves ~2× margin under the observed overflow point and absorbs per-step sync/header overhead for larger world sizes.

## Tests

Added 256 MiB / 512 MiB **aligned and unaligned** cases to `test_allgather_float16_{aligned,unaligned}`.

## Verification (RBLN-CR03, world_size=2, rebel-compiler 0.10.5.dev241)

- unaligned **32/32 OK** (incl. new 256/512 MiB)
- aligned **24/24 OK** (incl. new 256/512 MiB)
- previously-failing `numel 67109568` → **OK**

### DIFF
--- a/torch_rbln/csrc/distributed/c10d/rbln/ProcessGroupRBLN.cpp
+++ b/torch_rbln/csrc/distributed/c10d/rbln/ProcessGroupRBLN.cpp
@@ -67,8 +67,20 @@ constexpr size_t RCCL_ALLREDUCE_MAX_BYTES_PER_RANK = 1ULL * 1024ULL * 1024ULL;
 constexpr size_t RCCL_REDUCE_SCATTER_MAX_BYTES_PER_WORLD = 32ULL * 1024ULL * 1024ULL;
 constexpr size_t RCCL_MAX_BYTES_PER_REDUCE_SCATTER_OP = 2ULL * 1024ULL * 1024ULL;
 constexpr size_t RCCL_MAX_COMMAND_BUFFER_SUB_COMMANDS_COUNT = 15ULL;
-// AllGather size limit: output size <= 1GB (can be performed in a single allgather operation)
-constexpr size_t RCCL_ALLGATHER_MAX_OUTPUT_BYTES = 1ULL * 1024ULL * 1024ULL * 1024ULL;
+// AllGather per-call output-size limit (drives the chunked multi-call path).
+//
+// The REBEL/UMD AllGather builds its *entire* command stream into a single
+// fixed-size CS chunk (RCCL_CS_CHUNK_SIZE = 32 KiB) and ships exactly that many
+// bytes to the device. A transfer large enough that its per-page DMA
+// descriptors overflow 32 KiB has the tail descriptors silently dropped, so the
+// recv-buffer tail is left stale (or the op errors with -ENOMEM). Empirically
+// (world_size = 2) a 256 MiB output already overflows on REBEL while 128 MiB is
+// safe — see fsw-inference#324. CS size scales with the per-call *output* bytes,
+// so capping each AllGather call's output keeps the generated CS within budget;
+// anything larger is split into <= this many bytes per call by the chunked path
+// below. 64 MiB leaves ~2x margin under the observed overflow point and absorbs
+// the per-step sync/header overhead for larger world sizes.
+constexpr size_t RCCL_ALLGATHER_MAX_OUTPUT_BYTES = 64ULL * 1024ULL * 1024ULL;
 
 /**
  * @brief Get element size for RCCL operations based on dtype
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -95,12 +95,12 @@ def enable_eager_malloc(monkeypatch):
 # REBEL-failing tests keyed by fully expanded name -> (test file, reason). Keying on
 # the name pins one parametrization across separate ``@parametrize`` axes, which a
 # per-parameter mark cannot.
-_REBEL_XFAILS = {
-    "test_allgather_float16_unaligned_size_67109568_c10d_async_env_0_rbln": (
-        "test/distributed/test_process_group.py",
-        "REBEL: float16 unaligned all-gather fails at size 67109568 (64 MiB)",
-    ),
-}
+#
+# The float16 unaligned all-gather at size 67109568 (64 MiB) used to be xfailed
+# here; it is now fixed (AllGather is chunked to a CS-safe size, see
+# RCCL_ALLGATHER_MAX_OUTPUT_BYTES in ProcessGroupRBLN.cpp / fsw-inference#324),
+# so its strict-xfail entry was removed.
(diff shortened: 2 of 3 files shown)


rebel-chanheo (approved):
LGTM

torch_rbln/csrc/distributed/c10d/rbln/ProcessGroupRBLN.cpp
```diff
@@ -67,8 +67,20 @@ constexpr size_t RCCL_ALLREDUCE_MAX_BYTES_PER_RANK = 1ULL * 1024ULL * 1024ULL;
...
+// bytes to the device. A transfer large enough that its per-page DMA
+// descriptors overflow 32 KiB has the tail descriptors silently dropped, so the
+// recv-buffer tail is left stale (or the op errors with -ENOMEM). Empirically
+// (world_size = 2) a 256 MiB output already overflows on REBEL while 128 MiB is
+// safe — see fsw-inference#324. CS size scales with the per-call *output* bytes,
+// so capping each AllGather call's output keeps the generated CS within budget;
+// anything larger is split into <= this many bytes per call by the chunked path
+// below. 64 MiB leaves ~2x margin under the observed overflow point and absorbs
+// the per-step sync/header overhead for larger world sizes.
+constexpr size_t RCCL_ALLGATHER_MAX_OUTPUT_BYTES = 64ULL * 1024ULL * 1024ULL;
```
rebel-chanheo, line 83: Thanks for the detailed root-cause write-up — the chunking approach makes sense to me. One thing I'd like to confirm: are there side-effects we should be aware of from lowering `RCCL_ALLGATHER_MAX_OUTPUT_BYTES` from 1 GiB to 64 MiB?

1. Throughput for mid/large transfers — outputs in the 64 MiB–1 GiB range that used to complete in a single AllGather now take the chunked multi-call path, paying extra per-step sync/header overhead and more H2D round-trips.
2. Scaling with world_size — the 256 MiB overflow point was measured at world_size=2, but the CS also grows with the number of ring / recursive-doubling steps.
3. ATOM — the constant is device-agnostic, so ATOM (which didn't hit this bug) now also chunks at 64 MiB.

rebel-jongwonlee: Thanks — all three are fair, here's where each lands.

**Budget model used below.** CI confirms one AllGather's CS = a single 32 KiB chunk; the observed overflow is output 256 MiB @ world_size=2 (per-rank 128 MiB = 64 pages → local-copy 64 + send 64 + sync ≈ 130 descriptors). So the budget is ~128 descriptors of ~256 B. Data descriptors scale with **output** (≈ output/2 MiB, spread across the ring/RD sends); the per-step sync/header descriptors scale with **steps** and are independent of the output cap.

**1. Throughput, 64 MiB–1 GiB.** Splitting into (a) 256 MiB–1 GiB and (b) 64–256 MiB:
- (a) was **already broken** before this PR, not "completing in a single AllGather" — its CS overflowed 32 KiB, so the tail was dropped/stale (or -ENOMEM). Chunking it is a correctness fix, not a throughput regression.
- (b) genuinely worked single-call and now chunks. The cost is per-chunk CS build + a 32 KiB H2D (µs-scale) vs the actual data movement (ms-scale at these sizes) — well under ~1% wall-clock, and total bytes moved are unchanged. If we want to shrink even that, the world_size=2 "safe even when unaligned" point was 128 MiB output, so 128 MiB would halve the chunking for this band while still fixing the bug — happy to bump if you'd prefer that trade-off (it trades away some large-world margin, see #2).

**2. world_size scaling.** Two separate terms:
- *Data* descriptors stay ~constant under an output cap (per-rank = MAX/W shrinks as W grows), so capping output is actually the right knob here.
- *Sync* descriptors are ~4·(W−1), fixed per step and **independent of the output cap** — i.e. lowering the cap neither causes nor relieves them. At 64 MiB output the totals are ~36 desc @W=2 (~28% budget), ~44 @W=4, ~60 @W=8 (~47%) — a ~2× margin through CR03's 8-dev/node range. Beyond ~W≥24 the sync term alone approaches the 32 KiB chunk regardless of transfer size, but that's a pre-existing UMD limit this change doesn't move. Empirical anchor here is W=2; I kept 64 MiB (not the 128 MiB world=2 limit) specifically to hold margin across W≤8.

**3. ATOM.** The 32 KiB CS chunk (`RCCL_CS_CHUNK_SIZE`, rccl_util.h) is shared by both arches, so ATOM has the *same* latent CS-overflow ceiling — it just wasn't exercised by a large-enough unaligned case, not that it's immune. So chunking ATOM at 64 MiB is prudent, with the same small per-chunk overhead as #1(b). A device-aware cap (higher/disabled on ATOM) is doable but the rebel runtime headers expose no device-type query to this glue layer today, so it'd need new plumbing — I'd suggest keeping it device-agnostic for now and revisiting if ATOM throughput regresses in CI.

Net: I'd keep 64 MiB as-is (correctness + ~2× margin through W≤8). If you'd rather optimize the 64–256 MiB band, I'll move it to 128 MiB. Let me know which you prefer.

rebel-chanheo: Approve. Let's keep 64 MiB!

PR #97, opened 2026-06-18
fix: default RBLN control-plane IPs to loopback in rbln-ccl backend init

rebel-jonghewk:
## Summary of Changes

Default the rbln-ccl control-plane IPs (`RBLN_ROOT_IP` / `RBLN_LOCAL_IP`) on the production path so single-node distributed runs no longer die during process-group setup.

1. **Problem:** `apply_default_rbln_network_environment()` (which defaults these to `127.0.0.1`) only runs in autoport tests — never on a normal `init_process_group(backend="rbln-ccl")` path. So both vars stayed unset.
2. **Failure chain:** With them unset, rank 0's `rcclGetUniqueId` fails (`ExportMem` passes, failure only at unique_id), `broadcastUniqueId` throws `GetUniqueIdForBroadcast failed`, rank 0 dies, its rendezvous TCPStore closes, and every peer then dies with the misleading `c10d ... Failed to recv, got 0 bytes. Connection was likely closed` — making an IP-config problem look like a network/crash.
3. **Fix:** Call `_apply_control_plane_ips()` in `_create_process_group_rbln` before constructing `ProcessGroupRBLN`. Unset → `127.0.0.1`; pre-set values respected (multi-node); `RBLN_FORCE_LOOPBACK_CONTROL_PLANE=1` honored. Runs on every rank, so the rendezvous never starts half-configured.

---

## Related Issues / Tickets

* Related to # 

---

## Type of Change

- [x] `fix` — Bug fix

---

## Affected Modules

- [x] Backend

---

1. Single node, `RBLN_ROOT_IP` / `RBLN_LOCAL_IP` unset, run a TP>1 `rbln-ccl` workload (e.g. Qwen1.5-MoE TP=2).
2. Before: rank 0 logs `[Rccl] fail rcclGetUniqueId` + `GetUniqueIdForBroadcast failed`; peers cascade with `Failed to recv, got 0 bytes`.
3. After: both vars auto-default to `127.0.0.1`, `rcclGetUniqueId` succeeds, process group initializes.

---

Before (root cause):
```
[E] [runtime] [Rccl] fail rcclGetUniqueId (rank: -1, size: -1, ...)
[E] ProcessGroupRBLN: GetUniqueIdForBroadcast failed (call after PrepareContextAndExportMem)
... peers: torch.distributed.DistNetworkError: Failed to recv, got 0 bytes. Connection was likely closed.
```

---

## Notes

`_apply_control_plane_ips()` is an `_internal` helper; this wires the already-existing default logic into the live backend path instead of leaving it test-only. No native/C++ change — Python-only.

### DIFF
--- a/torch_rbln/__init__.py
+++ b/torch_rbln/__init__.py
@@ -165,6 +165,10 @@ def _create_process_group_rbln(dist_backend_opts, pg_options):
     """
     import torch_rbln._C
 
+    from torch_rbln._internal.rdma_env import _apply_control_plane_ips
+
+    _apply_control_plane_ips()
+
     # Extract parameters from dist_backend_opts
     store = dist_backend_opts.store
     group_rank = dist_backend_opts.group_rank
--- a/torch_rbln/__init__.py
+++ b/torch_rbln/__init__.py
@@ -34,8 +34,6 @@ def torch_backends_entry_point() -> None:
         import torch
 
         # Load shared objects ##################################################
-        global library_names, libraries
-
         # Ensure dependent shared library is loaded before importing native extension
         find_and_load_tvm_library("librbln.so")
 
@@ -164,7 +162,6 @@ def _create_process_group_rbln(dist_backend_opts, pg_options):
         ProcessGroupRBLN: A new ProcessGroupRBLN instance
     """
     import torch_rbln._C
-
     from torch_rbln._internal.rdma_env import _apply_control_plane_ips
 
     _apply_control_plane_ips()
--- a/test/distributed/test_control_plane_ip_default.py
+++ b/test/distributed/test_control_plane_ip_default.py
@@ -1,0 +1,68 @@
+# Owner(s): ["module: PrivateUse1"]
+
+import os
+
+import pytest
+import torch
+import torch.distributed as dist
+import torch.multiprocessing as mp
+
+from test.utils import configure_master_port_for_rccl_tests, requires_logical_devices, spawn_target_with_clean_exit
+
+
+def _run_allreduce_with_unset_control_plane_ips(rank: int, world_size: int, backend: str) -> None:
+    # The IPs must be unset on entry: this is the production path that the
+    # autoport helper masks by pre-setting them. Do NOT set them here.
+    assert not os.environ.get("RBLN_ROOT_IP"), "test precondition: RBLN_ROOT_IP must be unset"
+    assert not os.environ.get("RBLN_LOCAL_IP"), "test precondition: RBLN_LOCAL_IP must be unset"
+
+    os.environ["LOCAL_RANK"] = str(rank)
+    os.environ["WORLD_SIZE"] = str(world_size)
+    torch.rbln.set_device(rank)
+
+    dist.init_process_group(backend=backend, rank=rank, world_size=world_size)
+    try:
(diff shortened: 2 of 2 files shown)


torch_rbln/__init__.py
```diff
@@ -165,6 +165,10 @@ def _create_process_group_rbln(dist_backend_opts, pg_options):
     """
     import torch_rbln._C
 
+    from torch_rbln._internal.rdma_env import _apply_control_plane_ips
+
+    _apply_control_plane_ips()
```
rebel-chanheo, line 167: Could we add a regression test? The reason this slipped through is that the IP defaulting is only exercised on the autoport path, where `apply_default_rbln_network_environment()` pre-sets the IPs and therefore masks this exact bug. A test that reproduces the real production failure would both lock in the fix and serve as the empirical confirmation that rccl reads `RBLN_ROOT_IP`/`RBLN_LOCAL_IP` at process-group construction time (not at librbln load) — the test passing *is* that proof.

To actually reproduce the bug, the test should:
- **Not** call the autoport network helper (`configure_rbln_network_for_autoport_tests` / `apply_default_rbln_network_environment`) — it pre-sets the IPs and hides the regression.
- Run with `RBLN_ROOT_IP` / `RBLN_LOCAL_IP` **unset**.
- Single node, ≥2 logical devices (`@requires_logical_devices(2)`), under `test/distributed/`.
- Drive the real path: `init_process_group(backend="rbln-ccl")` **and** at least one actual collective (e.g. `all_reduce`/`barrier`), not just PG init — this also confirms the single-node data plane is fine with `RBLN_RDMA_IP` unset.
- Be marked `@pytest.mark.test_set_ci` and actually run in CI (not skipped on device count) — otherwise it's not a real guard.

rebel-jonghewk: Added in `test/distributed/test_control_plane_ip_default.py` (53dda18). It follows the requirements:

- Does **not** call the autoport helper; `RBLN_ROOT_IP`/`RBLN_LOCAL_IP` are `monkeypatch.delenv`'d and the child asserts they're unset on entry (so the test genuinely exercises the masked path).
- Single node, `@requires_logical_devices(2)`, under `test/distributed/`, `@pytest.mark.test_set_ci` so CI guards it.
- Drives the real path: `init_process_group(backend="rbln-ccl")` + an actual `all_reduce` (SUM), with `RBLN_RDMA_IP` unset — confirming the single-node data plane too.
- After init it asserts both IPs == `127.0.0.1`, i.e. the empirical proof that rccl reads them at PG-construction time, not at librbln load.

rebel-chanheo (approved):
LGTM

rebel-chanheo:
@rebel-jonghewk Is this a patch that should be included in the upcoming release? Please check.

rebel-jonghewk:
> @rebel-jonghewk Is this a patch that should be included in the upcoming release? Please check.

No, it's not a hard requirement.!

rebel-chanheo:
@rebel-jonghewk is this PR still good to merge? It looks like only the lint check is failing now. I can fix it if that would help.

rebel-jonghewk:
> @rebel-jonghewk is this PR still good to merge? It looks like only the lint check is failing now. I can fix it if that would help.

If you could push the lint fix, that'd be great, thanks!

PR #98, opened 2026-06-19
feat(copy): add native _foreach_copy_ op batching scatters into a single v2v submit

rebel-daeyang:
Related issue - https://github.com/rebellions-sw/vllm-rbln-internal/issues/65#issuecomment-4496432000

## Summary of Changes

Add a native RBLN implementation of `aten::_foreach_copy_` (`_foreach_copy__rbln`) and register it for the `PrivateUse1` backend.

`_foreach_copy_(self[], src[], non_blocking)` copies `src[i]` into `self[i]` for every `i`. The default decomposition lowers this to one `copy_` per pair, so a scatter into N separate-storage tensors (e.g. N per-layer KV caches) issues N independent runtime submits.

This op enqueues every eligible pair into a single `V2VBatch` and flushes them through **one** `rbln_memcpy_v2v_multi` submit — the write-side mirror of how `cat`/`stack` gather N sources into one destination through a single backend call. An N-way scatter collapses from N submits to 1.

Details:

- Conversion-free, same-device, same-shape pairs (`numel() > 0`) are enqueued via `strided_v2v_copy` into a shared `V2VBatch`, then submitted once.
- Pairs needing broadcast / dtype cast / cross-device transfer fall back to a plain per-pair `dst.copy_(src, non_blocking)`, so the general case is unchanged.
- `submit_or_fallback` preserves the existing CPU fallback path.

Files:

- `aten/src/ATen/native/rbln/RBLNCopy.{cpp,h}` — `_foreach_copy__rbln` impl + decl.
- `aten/src/ATen/native/RBLNRegisterOps.cpp` — register `_foreach_copy_` for `PrivateUse1`.

---

## Related Issues / Tickets

N/A

---

## Type of Change

- [ ] `fix` — Bug fix

---

## Affected Modules

- [ ] Backend
- [ ] Device
- [x] Ops/Kernels
- [ ] `torch.compile`
- [ ] Memory
- [ ] Build/Tools
- [ ] Docs

---

1. Build the extension against the matching rebel_compiler.
2. Run `_foreach_copy_` over a list of same-device, same-shape RBLN tensors and verify outputs match an equivalent per-pair `copy_` loop.
3. Edge cases: mixed shapes / dtypes / cross-device pairs fall back to per-pair `copy_` and stay correct; empty (`numel()==0`) tensors are skipped.

---

## Checklist

* [ ] Linting passes
* [ ] All CI tests pass

---

## Notes

Prerequisite for the LMCache-RBLN device-tensor `to_gpu` scatter speedup, which depends on this op plus a `set_device_layout_like` runtime API landing separately.

### DIFF
--- a/aten/src/ATen/native/RBLNRegisterOps.cpp
+++ b/aten/src/ATen/native/RBLNRegisterOps.cpp
@@ -135,6 +135,7 @@ TORCH_LIBRARY_IMPL(aten, PrivateUse1, m) {
   // Operations that use the device runtime API
   m.impl("_copy_from", TORCH_FN(at::native::rbln::_copy_from_rbln));
   m.impl("_copy_from_and_resize", TORCH_FN(at::native::rbln::_copy_from_and_resize_rbln));
+  m.impl("_foreach_copy_", TORCH_FN(at::native::rbln::_foreach_copy__rbln));
   m.impl("empty.memory_format", TORCH_FN(at::native::rbln::empty_rbln));
   m.impl("empty_strided", TORCH_FN(at::native::rbln::empty_strided_rbln));
   m.impl("zero_", TORCH_FN(at::native::rbln::zero_rbln_));
--- a/aten/src/ATen/native/rbln/RBLNCopy.cpp
+++ b/aten/src/ATen/native/rbln/RBLNCopy.cpp
@@ -320,6 +320,89 @@ at::Tensor clone_rbln(const at::Tensor& self, std::optional<c10::MemoryFormat> m
   return out;
 }
 
+namespace {
+
+// Byte extent [lo, hi) that `t` actually addresses, from sizes/strides — so
+// disjoint slices of one storage (e.g. kv_cache[0] vs kv_cache[1]) report
+// non-overlapping ranges. Negative strides extend `lo`; +1 makes it half-open.
+struct ByteRange {
+  const char* lo;
+  const char* hi;
+};
+
+ByteRange tensor_byte_range(const at::Tensor& t) {
+  const char* base = static_cast<const char*>(t.data_ptr());
+  const int64_t elsize = t.element_size();
+  int64_t lo_elems = 0;
+  int64_t hi_elems = 0;
+  const auto sizes = t.sizes();
+  const auto strides = t.strides();
+  for (size_t d = 0; d < sizes.size(); ++d) {
+    if (sizes[d] <= 1) {
+      continue;
+    }
+    const int64_t step = (sizes[d] - 1) * strides[d];
+    if (step >= 0) {
+      hi_elems += step;
+    } else {
+      lo_elems += step;
+    }
+  }
+  return {base + lo_elems * elsize, base + (hi_elems + 1) * elsize};
+}
+
+bool ranges_overlap(const ByteRange& a, const ByteRange& b) {
+  return a.lo < b.hi && b.lo < a.hi;
+}
+
+// The batched path submits eligible copies together (unordered) and after the
+// inline fallback copies, so it matches list-order copy_ only when copies don't
+// alias across pairs. Returns true on any cross-pair overlap — a destination
+// hitting another pair's source (RAW/WAR) or destination (WAW). Within-pair
+// (self[i]/src[i]) is left to copy_. Only same-device live tensors can alias.
+bool foreach_copy_reorder_unsafe(at::TensorList self, at::TensorList src) {
+  const size_t n = self.size();
+  std::vector<ByteRange> dst_range(n);
+  std::vector<ByteRange> src_range(n);
+  std::vector<bool> dst_live(n, false);
+  std::vector<bool> src_live(n, false);
+  std::vector<c10::DeviceIndex> dst_dev(n, -1);
+  std::vector<c10::DeviceIndex> src_dev(n, -1);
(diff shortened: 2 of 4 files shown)


aten/src/ATen/native/RBLNRegisterOps.cpp
```diff
@@ -135,6 +135,7 @@ TORCH_LIBRARY_IMPL(aten, PrivateUse1, m) {
   // Operations that use the device runtime API
   m.impl("_copy_from", TORCH_FN(at::native::rbln::_copy_from_rbln));
   m.impl("_copy_from_and_resize", TORCH_FN(at::native::rbln::_copy_from_and_resize_rbln));
+  m.impl("_foreach_copy_", TORCH_FN(at::native::rbln::_foreach_copy__rbln));
```
rebel-chanheo, line 138: Can we add tests for the new `_foreach_copy_` RBLN implementation?

This PR registers a new native backend op and changes copy behavior/performance in a fairly sensitive path, but I do not see direct coverage for it. It would be good to add tests for at least:

- same-device RBLN-to-RBLN batched copies
- mixed eligible/ineligible pairs, e.g. RBLN-to-RBLN plus CPU-to-RBLN
- non-contiguous source/destination views
- zero-numel tensors
- alias/order-sensitive cases where tensors appear in both `self` and `src`

The alias/order case is especially important because the implementation queues eligible copies and submits them after the loop, while fallback copies execute immediately. A regression test there would help ensure batching does not change `_foreach_copy_` semantics.

aten/src/ATen/native/rbln/RBLNCopy.cpp
```diff
@@ -318,4 +320,42 @@ at::Tensor clone_rbln(const at::Tensor& self, std::optional<c10::MemoryFormat> m
   return out;
 }
 
+void _foreach_copy__rbln(at::TensorList self, at::TensorList src, bool non_blocking) {
```
rebel-chanheo, line 406: I think this changes the observable behavior of `_foreach_copy_` when the tensor lists contain aliases and the copies are split between the batched V2V path and the immediate fallback path.

In `_foreach_copy__rbln`, eligible RBLN-to-RBLN pairs are only enqueued into `V2VBatch` and are submitted after the loop, while ineligible pairs are executed immediately via `dst.copy_(s, non_blocking)`. This can let a later fallback copy mutate the source tensor of an earlier queued copy before the queued copy is actually submitted.

A minimal repro would look like this:

```python
import torch
import torch_rbln

device = "rbln:0"

a = torch.tensor([1.0, 2.0], device=device)
b = torch.tensor([3.0, 4.0], device=device)
cpu_src = torch.tensor([9.0, 10.0])  # CPU source forces the fallback path for b <- cpu_src

torch._foreach_copy_([a, b], [b, cpu_src])

print(a.cpu())
print(b.cpu())
```

Expected behavior, matching pairwise/list-order copy semantics:

a == tensor([3., 4.])    # copied from the original b
b == tensor([9., 10.])

With the current batching logic, the execution order becomes effectively:

enqueue a <- b   # not executed yet
execute b <- cpu_src immediately
submit a <- b    # now reads the already-mutated b

So a can incorrectly become [9., 10.].

I think we need to preserve ordering when batched and non-batched copies are mixed. Possible fixes are to flush the current batch before executing an immediate fallback copy, or to detect cross-list alias/overlap and fall back to a fully sequential copy_ path for that case. A regression test for this mixed eligible/ineligible alias case would be useful.

rebel-daeyang: Thanks for the good catch — confirmed and fixed. The hazard isn't limited to the batched-vs-fallback split you noted: two batch-eligible pairs that alias (e.g. _foreach_copy_([a, b], [b, a])) are also reordered, since the V2V multi-copy has no inter-copy ordering. So I thought flushing before each fallback isn't enough.

Instead, before the loop I check for cross-pair overlap — any destination that overlaps another pair's source (RAW/WAR) or destination (WAW) — and fall back to a strictly sequential copy_ loop when found; otherwise the batched fast path runs unchanged. The check uses actual byte ranges (from sizes/strides), not storage identity, so disjoint views of one storage (e.g. kv_cache[0] / kv_cache[1]) are correctly treated as non-overlapping and stay on the fast path. The common scatter has no overlap, so it's free in practice.

Added a regression test for your exact reproduction case plus the swap-within-batch and overlapping-destination cases.
I also measured `foreach_copy_reorder_unsafe` — it runs in ~10–30µs, so it adds no meaningful overhead to the fast path.

aten/src/ATen/native/RBLNRegisterOps.cpp
rebel-daeyang: Added `test/rbln/test_foreach_copy_v2v.py` covering all of these:
* same-device batched copies (incl. many pairs and all engine dtypes),
* mixed eligible/ineligible (RBLN + CPU source, dtype-cast fallback),
* non-contiguous source/destination views,
* zero-numel pairs,
* and the alias/order-sensitive cases — your [a, b] <- [b, cpu_src] reproduction, the all-eligible swap,
* and overlapping destinations (WAW).

Each compares against the list-order CPU reference.

rebel-chanheo (approved):
LGTM

PR #99, opened 2026-06-19
feat(memory): add set_device_layout_like to match a staging buffer's layout to a device tensor

rebel-daeyang:
## Summary of Changes

Add `set_device_layout_like(target, ref)` to the RBLN memory API: configure `target`'s device allocation to the same physical layout (NT/WT kind + dtype) as a device-resident `ref` tensor, **without copying any data**.

A device-to-device copy only takes the runtime fast path when both sides share the same layout kind. By making a freshly allocated staging buffer adopt a KV cache's (with-transform) layout up front, the bulk host→device staging fill and the per-slot D2D scatter that follows both stay on the fast path. `ref` must already be device-resident (have a physical view); `target` keeps its own size and only mirrors the layout kind and dtype.

The new API is exposed at three layers:

- `c10::rbln::set_device_layout_like(void* target_data, const void* ref_data)` — C++ entry point that forwards to the `rbln_set_device_alloc_layout_like` runtime API.
- `_set_device_layout_like(target, ref)` — internal pybind taking two `at::Tensor`s.
- `torch_rbln.set_device_layout_like(target, ref)` — Python wrapper.

Files:

- `c10/rbln/RBLNFunctions.{cpp,h}` — `set_device_layout_like` impl + declaration.
- `torch_rbln/csrc/rbln/Module.cpp` — `_set_device_layout_like` pybind registration.
- `torch_rbln/memory.py` — `torch_rbln.set_device_layout_like` wrapper.

---

## Related Issues / Tickets
https://github.com/rebellions-sw/vllm-rbln-internal/issues/65#issuecomment-4496432000 - LMCache-RBLN에 device tensor 적용
https://github.com/rebellions-sw/rebel_compiler/pull/11610 - update(vmemory): add set_device_alloc_layout_like runtime API

---

## Type of Change

- [ ] `fix` — Bug fix

---

## Affected Modules

- [ ] Backend
- [ ] Device
- [ ] Ops/Kernels
- [ ] `torch.compile`
- [x] Memory
- [ ] Build/Tools
- [ ] Docs

---

1. Build the extension against a rebel_compiler that exposes `rbln_set_device_alloc_layout_like` (see Notes).
2. Allocate a device-resident `ref` tensor and a fresh `target` staging buffer, call `torch_rbln.set_device_layout_like(target, ref)`, then run a host→`target` fill followed by a `target`→`ref` device-to-device copy and confirm it stays on the fast path (no per-op host bounce).
3. Edge cases: passing a `ref` without a physical view, or a null tensor, raises.

---

## Checklist

* [ ] Linting passes
* [ ] All CI tests pass

---

## Notes

Depends on the `rbln_set_device_alloc_layout_like` runtime API in rebel_compiler, which lands in a separate PR (https://github.com/rebellions-sw/rebel_compiler/pull/11610). This branch will not build until that rebel_compiler change is available. Together with `_foreach_copy_` (#98), this is a building block for the LMCache-RBLN device-tensor `to_gpu` scatter speedup.


c10/rbln/RBLNFunctions.cpp
```diff
@@ -270,6 +270,16 @@ void free(void* data) {
   RBLN_CHECK(!::rbln::rbln_free(vaddr));
 }
 
+void set_device_layout_like(void* target_data, const void* ref_data) {
+  RBLN_CHECK(target_data != nullptr, "set_device_layout_like: target is nullptr");
+  RBLN_CHECK(ref_data != nullptr, "set_device_layout_like: ref is nullptr");
+  const auto target_vaddr = reinterpret_cast<uint64_t>(target_data);
+  const auto ref_vaddr = reinterpret_cast<uint64_t>(ref_data);
+  RBLN_LOG_DEBUG("set_device_layout_like: target={:#x} ref={:#x}", target_vaddr, ref_vaddr);
+  RBLN_CHECK(!::rbln::rbln_set_device_alloc_layout_like(target_vaddr, ref_vaddr),
```
rebel-chanheo: This PR calls `::rbln::rbln_set_device_alloc_layout_like(...)`, but the function is not declared in the vendored runtime header at `third_party/rebel_compiler/include/rebel/runtime/api/rbln_runtime_api.h`.

Since the default build uses the vendored header, this will likely fail to compile unless the header is updated as part of this PR. If this depends on a newer `rebel-compiler`, please also update the vendored header and the required/pinned compiler version accordingly.

torch_rbln/memory.py
```diff
@@ -46,6 +47,23 @@ def _normalize_device(device: Optional[Union[int, str, torch.device]]) -> torch.
         return device
 
 
+def set_device_layout_like(target: torch.Tensor, ref: torch.Tensor) -> None:
```
rebel-chanheo, line 50: This is exposed as a `torch.rbln` API, so I think we should validate the tensor inputs before passing raw `data_ptr()` values to the runtime.

Right now `_set_device_layout_like` accepts any `at::Tensor` and immediately calls:

```cpp
c10::rbln::set_device_layout_like(target.data_ptr(), ref.data_ptr());
```

If a user accidentally passes a CPU tensor, a tensor on another backend, or an incompatible tensor, we may pass a host pointer to a runtime function that expects an RBLN vaddr. That would likely produce a low-level runtime failure instead of a clear user-facing error.

Can we add explicit checks before calling the runtime? For example:

```
TORCH_CHECK(target.device().is_privateuseone(), "target must be an RBLN tensor");
TORCH_CHECK(ref.device().is_privateuseone(), "ref must be an RBLN tensor");
TORCH_CHECK(target.scalar_type() == ref.scalar_type(), "target and ref must have the same dtype");
```

If there are additional runtime requirements, such as ref needing an existing physical view or both tensors needing compatible device/layout state, those should also be checked where possible and surfaced with clear error messages. A small test for invalid CPU/non-RBLN inputs would be useful too.

torch_rbln/memory.py
```diff
@@ -15,6 +15,7 @@
 
 __all__ = [
     "empty_cache",
+    "set_device_layout_like",
```
rebel-chanheo, line 18: Can we add tests for this new API?

Since `set_device_layout_like` is exposed through `torch.rbln`, it would be useful to cover both API visibility and invalid input handling. Suggested cases:

- `hasattr(torch.rbln, "set_device_layout_like")`
- valid RBLN tensor inputs call the binding successfully
- CPU tensor as `target` raises a clear error
- CPU tensor as `ref` raises a clear error
- mismatched dtype or otherwise incompatible tensor metadata raises a clear error, if that is not supported

These tests would help ensure users get actionable errors instead of low-level runtime failures from passing invalid pointers or incompatible tensors into the runtime.

c10/rbln/RBLNFunctions.cpp
rebel-daeyang: Agreed, this is gated on the runtime symbol shipping. The vendored header decl only works once librbln.so actually exports it, so I'll add the declaration together with the rebel-compiler version bump after rebellions-sw/rebel_compiler#11610 is released. Until then the branch builds against an external rebel_compiler.

torch_rbln/memory.py
rebel-daeyang: Added checks in the binding before any `data_ptr()` reaches the runtime: both must be RBLN tensors, same device, same dtype. The dtype check is required, not just defensive — the runtime computes the element count as target_bytes / ref_elem_size and stamps ref's dtype onto target's alloc, so a differing dtype truncates the count or silently re-types the buffer (garbage on readback). The "ref needs a physical view" precondition is enforced by the runtime (currently a generic failure message; a clearer one would need a runtime-API change).

torch_rbln/memory.py
rebel-daeyang: Added `test/rbln/test_set_device_layout_like.py`: API visibility, CPU target / CPU ref / dtype-mismatch all raise with clear messages, and a ref without a physical view raises. The successful end-to-end call needs a device-resident KV ref, which a unit test can't materialize, so that path is covered by the LMCache-RBLN integration.

torch_rbln/csrc/rbln/Module.cpp
```diff
@@ -112,6 +112,30 @@ void register_internal_api(py::module_& module) {
...
 
+  // Set target tensor's device-alloc layout (NT/WT kind + dtype) to match ref,
+  // without copying data. Used by torch_rbln.set_device_layout_like().
+  module.def(
+      "_set_device_layout_like",
+      [](const at::Tensor& target, const at::Tensor& ref) {
+        // Validate before passing raw data_ptr()s to the runtime (which expects
+        // RBLN vaddrs). dtype must match: the runtime sizes target's alloc by
+        // ref's element size, so a mismatch re-types the buffer under target.
+        TORCH_CHECK(
```
rebel-chanheo: The device / same-device / dtype checks here look great. One remaining input that slips through: a view/slice target or ref with a non-zero storage_offset.

data_ptr() returns storage_base + storage_offset * itemsize, so a view hands a non-base vaddr to the runtime. The runtime looks it up by exact match, so a view tensor fails with a cryptic `Invalid target_vaddr=...` rather than a clear, actionable error. And even for an offset-0 narrowing view, the runtime sizes the layout from the whole allocation (`target_it->second->size()`), not the tensor's logical nbytes, so it would be misconfigured.

Since this is a public torch.rbln API, could we also require a base allocation, alongside the existing checks?

```
TORCH_CHECK(
    target.storage_offset() == 0 && target.is_contiguous(),
    "set_device_layout_like: target must be a base (non-view) RBLN tensor");
TORCH_CHECK(
    ref.storage_offset() == 0 && ref.is_contiguous(),
    "set_device_layout_like: ref must be a base (non-view) RBLN tensor");
```

rebel-daeyang: Good point — fixed. Added a base-allocation check for both target and ref.

I made it a bit stricter than the snippet: `storage_offset()==0 && is_contiguous()` still admits an offset-0 narrowing view like base[:k] (it's contiguous, so it would slip through and be mis-handled), so I also require the tensor to span its whole storage.

Added tests for offset>0 and offset-0 narrowing views (both target and ref), and documented the base-allocation / dtype / same-device requirements in the docstring.

rebel-chanheo (approved):
LGTM

rebel-daeyang:
@rebel-chanheo Should the `constraints-build-dev.txt` pin always point to a .prod build, or is a plain dev wheel fine?

PR #100, opened 2026-06-19
feat(rbln): allow device_count() == 0 for compile-only use without an NPU

rebel-chanheo:
## Summary of Changes

Treat `device_count() == 0` as a valid, non-fatal state (the CUDA model) so a sharded model can be traced/compiled on a host with no NPU.

Previously the device-mapping initializer asserted at least one physical NPU, so importing torch-rbln or building a DTensor device mesh (which constructs a `ProcessGroupRBLN`) crashed on a host with no hardware (`No logical device (this process has 0 physical NPU(s) ...)`).

Now:
- Device enumeration reports `0` instead of throwing — like `torch.cuda.device_count()` on a CPU-only host; `is_available()` / `hasRBLN()` report unavailable.
- Selecting a current logical device is bookkeeping and is allowed with 0 devices, so a device mesh can place tensors on `rbln:0..N-1` during tracing.
- A `world_size > 1` `ProcessGroupRBLN` constructs without hardware (RCCL init is skipped when there are 0 devices).
- Every operation that actually touches a device — allocation, `synchronize`, `empty_cache`, memory stats, storage resize — fails cleanly at the shared logical→physical resolution chokepoint with a clear message, rather than at init or deep in the runtime.

A failed device-count *query* (SDK / driver not loadable) stays fatal — "cannot determine the count" is distinct from "the count is 0".

## Related Issues / Tickets

* Related to # 

## Type of Change

## Affected Modules

- [x] Backend
- [x] Device

1. On a host with **no NPU**: `pytest test/rbln/test_no_device.py` and `./build/bin/test_cpp_RBLNNoDeviceTest` — the full 0-device contract runs (import + `device_count() == 0`, `set_device` bookkeeping, allocation/synchronize/empty_cache fail cleanly, `ProcessGroupRBLN(size=2)` constructs).
2. On a host with NPUs the 0-device contract tests skip; `test/rbln/test_device_mapping.py` and `RBLNFunctionsTest` confirm the production strict path is unchanged.

## Checklist

* [x] Linting passes
* [ ] All CI tests pass

## Notes

The guard lives at the single `to_device_id()` chokepoint that all index-based device operations share, so the 0-device behavior is uniform rather than per-call.

### DIFF
--- a/c10/rbln/DeviceMappingManager.cpp
+++ b/c10/rbln/DeviceMappingManager.cpp
@@ -281,9 +281,23 @@ void DeviceMappingManager::initialize() {
     RBLN_LOG_DEBUG("Initializing RBLN device mapping");
 
     int device_count = 0;
+    // A failed query (SDK/driver not loadable) stays fatal; "query succeeded,
+    // found 0 NPUs" is handled below.
     RBLN_CHECK(!rbln_get_device_count(&device_count));
     const int physical_device_count = device_count;
     RBLN_LOG_DEBUG("Found {} physical NPU(s)", physical_device_count);
+
+    // No physical NPU: register 0 logical devices instead of failing (like
+    // torch.cuda.device_count()==0 on a CPU-only host). Device use fails at the
+    // point of use, so a model can still be traced/compiled without an NPU.
+    if (physical_device_count == 0) {
+      RBLN_LOG_INFO(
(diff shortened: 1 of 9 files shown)


rebel-jonghewk (commented):
lgtm

PR #111, opened 2026-06-25
fix(rbln): revert weight-free compilation default (RBLN_WEIGHT_FREE=on)

rebel-chanheo:
## What

Reverts #90 (commit `cfbab22`), which made weight-free compilation the default by forcing `RBLN_WEIGHT_FREE=on` when the user hadn't set it.

Restores the previous behavior: `RBLN_WEIGHT_FREE` is left unset by default; users can still opt in explicitly with `RBLN_WEIGHT_FREE=on`.

## Why

Defaulting to `on` surfaced a stability issue — fp16 graph-mode `inf`/`nan` under weight-free compilation. Until that's fixed in the compiler, the safe default is to not force weight-free on for everyone.

## Scope of the revert

- `torch_rbln/__init__.py` — remove `_apply_weight_free_default()`, its call in `torch_backends_entry_point()`, and the now-unused `rbln_log_info` import.
- `test/rbln/test_weight_free.py` — remove the regression tests added alongside the feature.
- **`constraints-build-dev.txt` is intentionally NOT reverted.** #90 bumped rebel-compiler `dev269 → dev337`; the file is now at `dev407` on `main`. Reverting it would downgrade the compiler, so it's kept at the current `main` version. Only the weight-free **default behavior** is reverted.

## Checks

- `ruff check` ✅
- `ruff format --check` ✅
- `mypy` ✅
- No dangling references to the removed symbols.


rebel-jonghewk (approved):
Thanks. let's revisit this after merging https://github.com/rebellions-sw/rebel_compiler/pull/11680

rebel-clee (approved):
LGTM

PR #116, opened 2026-06-29
feat(profiler): add torch.profiler support via libkineto bridge

rebel-ywkim:
## Summary of Changes

This PR adds **`torch.profiler` support for the RBLN NPU backend** by implementing the
consumer side of the rbln ↔ libkineto bridge.

**What this PR does.** It registers an RBLN `IActivityProfiler` with libkineto. When a
`torch.profiler` session runs, the plugin drives rebel_compiler's `rbln_kineto_*` C-ABI to
pull the materialized rbln trace and converts it into libkineto activities, so the RBLN
device/host timeline appears in the **same chrome trace** as the Python/CPU events.

**Added:**

- **libkineto plugin (adapter)** — `torch_rbln/csrc/rbln/profiler/kineto/rbln_kineto_adapter.{cc,h}`
  An `IActivityProfiler` / `IActivityProfilerSession` whose `start()`/`stop()` drive the C-ABI
  (`rbln_kineto_begin_session` / `rbln_kineto_end_session_and_export`), gated on
  `rbln_kineto_is_active()` so a session is created only when rbln profiling is actually on.
- **trace projection (emitter)** — `torch_rbln/csrc/rbln/profiler/kineto/rbln_kineto_emitter.{cc,h}`
  Converts the exported rbln POD (`RblnKinetoExport`: devices / lanes / slices) into libkineto
  `DeviceInfo` / `ResourceInfo` / `GenericTraceActivity`, applying the steady→system clock offset.
- **registration** — registers the profiler factory with libkineto at `_C` init.
- **vendored C-ABI headers** — `third_party/rebel_compiler/include/rebel/runtime/api/rbln_kineto_api.h`
  (+ `rbln_retcode.h`), the contract shared with rebel_compiler.

**Expected behavior**

| Setup | Result |
|---|---|
| `RBLN_PROFILER=1` **and** `torch.profiler` | Python profiling data within the scope **and** RBLN runtime profiling are merged into one chrome trace. The Query Profiler still writes its `.pb` for the events inside the scope. |
| `RBLN_PROFILER=1` only (or torch-rbln without this kineto plugin) | Unchanged — standalone `.pb` from the Query Profiler, independent of torch. |
| `torch.profiler` only (no `RBLN_PROFILER`) | Unchanged — default chrome trace, with no RBLN profiler data. |

---

---
## Related Issues / Tickets

* https://github.com/rebellions-sw/rebel_compiler/pull/11658

---

## Type of Change

- [ ] `fix` — Bug fix

---

## Affected Modules

- [ ] Backend
- [ ] Device
- [ ] Ops/Kernels
- [ ] `torch.compile`
- [ ] Memory
- [x] Build/Tools
- [ ] Docs

---

## How to Test

> Requires torch-rbln built against a rebel_compiler that exports the `rbln_kineto_*`
> (session) C-ABI — see Notes.

1. Profile an RBLN model under both flags:

   ```python
   import torch   # registers the libkineto profiler
   from torch.profiler import profile
   # ... load / compile a model on the RBLN backend ...
   with profile(with_stack=True) as prof:
       model(inputs)            # runs on the RBLN NPU
   prof.export_chrome_trace("trace.json")
   ```

   ```bash
   RBLN_PROFILER=1 RBLN_PROFILER_DIR=./rbln_prof python run_profile.py
   ```

2. Open `trace.json` (chrome://tracing or ui.perfetto.dev) and verify RBLN device rows
   (compute / DMA / sync slices) appear **alongside** the Python/CPU events.
3. Verify `./rbln_prof/session_<N>/rbln_*.pb` is still produced (Query Profiler intact).
4. Negative checks:
   - `RBLN_PROFILER=1` alone (no `torch.profiler`) → only the `.pb`; no behavior change.
   - `torch.profiler` alone (no `RBLN_PROFILER`) → default chrome trace; no RBLN rows.

Expected result: device/compute/DMA slices from the rbln trace render under their own rows
in the merged chrome trace, with timestamps aligned to the Python timeline.

---

---

## Checklist

---

## Notes

---

### DIFF
--- a/third_party/rebel_compiler/include/rebel/runtime/api/rbln_runtime_api.h
+++ b/third_party/rebel_compiler/include/rebel/runtime/api/rbln_runtime_api.h
@@ -7,7 +7,7 @@
 #include <tuple>
 #include <vector>
 
-#include "rebel/runtime/api/rbln_retcode.h"
+#include <rebel/runtime/api/rbln_retcode.h>
 
 #ifdef __cplusplus
 
--- a/third_party/rebel_compiler/include/rebel/runtime/api/rbln_runtime_api.h
+++ b/third_party/rebel_compiler/include/rebel/runtime/api/rbln_runtime_api.h
@@ -1,13 +1,12 @@
 #ifndef REBEL_RUNTIME_API_RBLN_RUNTIME_API_H
 #define REBEL_RUNTIME_API_RBLN_RUNTIME_API_H
 
+#include <rebel/runtime/api/rbln_retcode.h>
 #include <stdint.h>
 
 #include <string>
 #include <tuple>
 #include <vector>
-
-#include <rebel/runtime/api/rbln_retcode.h>
 
 #ifdef __cplusplus
 
--- a/third_party/rebel_compiler/include/rebel/runtime/api/rbln_runtime_api.h
+++ b/third_party/rebel_compiler/include/rebel/runtime/api/rbln_runtime_api.h
@@ -7,6 +7,8 @@
 #include <tuple>
 #include <vector>
 
+#include "rebel/runtime/api/rbln_retcode.h"
+
 #ifdef __cplusplus
 
 namespace rbln {
@@ -15,12 +17,6 @@ class MemoryStats;
 
 extern "C" {
 #endif
-
-typedef enum {
-  RBLNRetCode_SUCCESS = 0,
-  RBLNRetCode_FAILURE,
-  RBLNRetCode_INVALID,
-} RBLNRetCode;
 
 typedef enum {
   RBLNMemcpyType_H2D = 0,
@@ -278,15 +274,10 @@ constexpr uint32_t kMaxV2VMultiCopies = 1024;
 RBLNRetCode rbln_memcpy_v2v_multi(
     const std::vector<std::tuple<uint64_t, uint64_t, uint64_t>>& copies);
 
-// Async variants. *handle_out == 0 means a synchronous (fast-path-ineligible) completion;
-// else rbln_transfer_wait before reading dst. Sync copy APIs don't chain on the stream —
-// wait/synchronize before a sync copy on the same vaddr.
-//
-// Ordering contract: between an async dispatch and its rbln_transfer_wait, do NOT issue any
-// other copy that touches the SAME device area (not just the same vaddr — distinct vaddrs in
-// one vmem entry share its area, hence the same cached command buffer). The finalize step
-// reads the area's MRU command-buffer entry, which an intervening copy would rotate, so an
-// unaligned D2H bounce would land at the wrong host address.
+// Async variants. *handle_out == 0 means a synchronous (fast-path-ineligible) completion; else
+// rbln_transfer_wait (or rbln_device_synchronize) before reading dst — there is no auto-sync on
+// host read. The same vaddr/area may be re-copied while a transfer is in flight (finalize is
+// per-handle, not MRU-order). Sync copy APIs drain the device's in-flight transfers first.
 RBLNRetCode rbln_memcpy_v2h_async(uint64_t src_vaddr, uintptr_t dst_host_ptr, uint64_t size,
                                   uint64_t* handle_out);
 RBLNRetCode rbln_memcpy_h2v_async(uintptr_t src_host_ptr, uint64_t dst_vaddr, uint64_t size,
--- a/torch_rbln/csrc/rbln/profiler/kineto/rbln_kineto_adapter.cc
+++ b/torch_rbln/csrc/rbln/profiler/kineto/rbln_kineto_adapter.cc
@@ -4,6 +4,8 @@
 
 #include <torch_rbln/csrc/rbln/profiler/kineto/rbln_kineto_adapter.h>
 
+#include <c10/rbln/RBLNLogging.h>
+
 #include <kineto/ActivityType.h>
 #include <kineto/Config.h>
 #include <kineto/GenericTraceActivity.h>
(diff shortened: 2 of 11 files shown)


third_party/rebel_compiler/include/rebel/runtime/api/rbln_runtime_api.h
```diff
@@ -7,6 +7,8 @@
 #include <tuple>
 #include <vector>
 
+#include "rebel/runtime/api/rbln_retcode.h"
```
rebel-ysseo: Please change this part from “” (double quotation marks) to <> (angle brackets).

torch_rbln/csrc/rbln/profiler/kineto/rbln_kineto_adapter.cc
```diff
@@ -0,0 +1,176 @@
...
+    return {};
+  }
+
+  std::unique_ptr<::libkineto::CpuTraceBuffer> getTraceBuffer() override {
+    auto buf = std::make_unique<::libkineto::CpuTraceBuffer>();
+    for (const auto& act : projected_.activities) {
+      buf->emplace_activity(act);
+    }
+    return buf;
+  }
```
rebel-jonghewk, line 127: What are these for?

torch_rbln/csrc/rbln/profiler/kineto/rbln_kineto_adapter.cc
```diff
@@ -0,0 +1,176 @@
...
+  RblnActivityProfilerSession& operator=(const RblnActivityProfilerSession&) = delete;
+
+  void start() override {
+    status_ = ::libkineto::TraceStatus::RECORDING;
+    // (steady, system) anchor at session open; exported slices carry steady_clock ns,
+    // so we add (system - steady) at projection time to convert them to system time.
+    anchor_steady_ns_ = now_ns(std::chrono::steady_clock::now());
+    anchor_system_ns_ = now_ns(std::chrono::system_clock::now());
+    start_ts_ns_ = anchor_system_ns_;
+    rbln_kineto_begin_session();
```
rebel-jonghewk: Is it okay to ignore the return code here?

torch_rbln/csrc/rbln/profiler/kineto/rbln_kineto_adapter.cc
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  void stop() override {
+    status_ = ::libkineto::TraceStatus::PROCESSING;
+    clock_offset_ns_ = anchor_system_ns_ - anchor_steady_ns_;
+    projected_ = ProjectedKinetoTrace{};
+    int32_t exported = 0;
+    rbln_kineto_end_session_and_export(&sink_thunk, this, &exported);
```
rebel-jonghewk: Same here.

rebel-jonghewk:
@rebel-ywkim I left a few comments — none are blockers, but they're worth noting.

torch_rbln/csrc/rbln/profiler/kineto/rbln_kineto_adapter.cc
```diff
@@ -0,0 +1,176 @@
...
+    // so we add (system - steady) at projection time to convert them to system time.
+    anchor_steady_ns_ = now_ns(std::chrono::steady_clock::now());
+    anchor_system_ns_ = now_ns(std::chrono::system_clock::now());
+    start_ts_ns_ = anchor_system_ns_;
+    rbln_kineto_begin_session();
+  }
+
+  void stop() override {
+    status_ = ::libkineto::TraceStatus::PROCESSING;
+    clock_offset_ns_ = anchor_system_ns_ - anchor_steady_ns_;
```
rebel-ysseo: `anchor_system_ns_` / `anchor_steady_ns_` are only set in `start()` and never updated after, so why compute `clock_offset_ns_` in `stop()` instead of in `start()`?

torch_rbln/csrc/rbln/profiler/kineto/rbln_kineto_adapter.cc
rebel-ywkim: This is one of the virtual function for IActivityProfilerSession needed to be implemented. 

In kineto side, `processTraceInternal` handles logging.  (https://github.com/pytorch/kineto/blob/8edb29efc1ed18255be3daa924580c88c8add76a/libkineto/src/GenericActivityProfiler.cpp#L211 ),
I added the logic that copies all the data into the logger to `processTrace()`, thereby making `finalizeTrace()`, which calls `getDeviceInfo` , `getResourceInfos` and `getTraceBuffer` does not actually affect the overall result of Chrometrace. ( `getDeviceInfo`, `getResourceInfos` I already left as no-op ).
`getTraceBuffer` is still called inside libkineto's `processTraceInternal` but the transferred activity will be ignored as further code does not refer to pushed activities.

so basically `getTraceBuffer()` is currently no-op in terms of chrometrace generation, and we can make this function as no-op also. However, I just left this function to copy activities following some references.

torch_rbln/csrc/rbln/profiler/kineto/rbln_kineto_adapter.cc
```diff
@@ -0,0 +1,176 @@
...
+
+  void on_export(const RblnKinetoExport* exp) {
+    convert_export_to_kineto(exp, clock_offset_ns_, default_trace_span(), &projected_);
+  }
+
+  std::vector<std::string> errors() override {
+    return {};
+  }
+
+  void processTrace(::libkineto::ActivityLogger& logger) override {
```
rebel-chanheo, line 102: Could we avoid emitting the same activities through both paths? Kineto processes child sessions via processTrace(...), and later also collects getTraceBuffer() during finalization. Since processTrace() already logs projected_.activities directly, returning the same activities from getTraceBuffer() may duplicate events or expose them as CPU trace buffers. If this profiler owns logging through processTrace(), returning nullptr or an empty buffer here seems safer.

torch_rbln/csrc/rbln/profiler/kineto/rbln_kineto_emitter.cc
```diff
@@ -0,0 +1,98 @@
...
+  out->activities.reserve(exp->slices_count);
+  for (uint32_t i = 0; i < exp->slices_count; ++i) {
+    const RblnKinetoSlice& s = exp->slices[i];
+    const ::libkineto::ActivityType act_type = kind_to_activity_type(s.kind);
+    ::libkineto::GenericTraceActivity act(span, act_type, std::string(s.name ? s.name : ""));
+    act.startTime = s.start_steady_ns + clock_offset_ns;
+    act.endTime = s.end_steady_ns + clock_offset_ns;
+    act.device = s.device_pid;
+    act.resource = s.resource_tid;
+    act.id = static_cast<int32_t>(s.corr_id);
```
rebel-chanheo: Could we avoid the unchecked narrowing of corr_id here? The C ABI defines `corr_id` as int64_t, but `GenericTraceActivity::id` is int32_t, so values outside the 32-bit range can truncate and collide. We should either constrain the ABI to 32-bit correlation IDs or add a checked mapping / preserve the original value in metadata.

torch_rbln/csrc/rbln/profiler/kineto/rbln_kineto_adapter.cc
rebel-ywkim: I agree, since I added logging logics into the processTrace(), It would be more natural and safer to leave getTraceBuffer as no op. 

cc. @rebel-jonghewk

torch_rbln/csrc/rbln/profiler/kineto/rbln_kineto_adapter.cc
rebel-ywkim: I will add retcode checking logic and make it gracefully return with `RBLN_LOG_WARN` once rbln kineto api fails, will it be okay ?

torch_rbln/csrc/rbln/profiler/kineto/rbln_kineto_emitter.cc
rebel-ywkim: corr_id was introduced as the interface for integrating internally managed 64bit IDs with kineto for correlation later. However, correlation is currently unimplemented. 
So can I drop corr_id from both sides for now, and re-introduce it together with the correlation feature when it's actually implemented?

rebel-chanheo: Sounds good to me. If correlation is not implemented yet, dropping corr_id for now is cleaner than keeping an ABI field that we cannot preserve correctly in Kineto.

rebel-chanheo (approved):
LGTM!

rebel-jonghewk (approved):

rebel-ysseo (approved):

PR #120, opened 2026-06-30
feat(device): dummy device (RBLN_DUMMY_DEVICE) for no-NPU compile

rebel-chanheo:
## What

`RBLN_DUMMY_DEVICE=1` lets torch-rbln build and compile a model with no NPU present (compile-only on a host without hardware). Allocations and copies are backed by rebel's v-memory host mirror through a dummy runtime context — torch adds no host-memory shim of its own.

## torch-side changes (minimal)

- Detect `RBLN_DUMMY_DEVICE` and register the logical device(s) via the same `registerLogicalDevice` the real path uses, so allocation/copy flow through v-memory unchanged. Layout comes from `RBLN_DEVICE_MAP` / `RBLN_NPUS_PER_DEVICE` (ids are shape markers, not range-checked against hardware).
- `physical_device_count()` reports 0 without querying the runtime.
- `set_device_layout_like()` is a no-op — the only device-only op with no host-side meaning.
- Executing a compiled graph raises a clear error (it is compile-only; the dummy runtime would otherwise silently return zeros). `mode=compile_only` builds are allowed through.
- `is_dummy_device()` exposed to Python.

No new allocation/copy path; every dummy branch is gated by a cached bool, so real-device behavior is unchanged.

## Setup

`RBLN_TARGET_SOC` must be set alongside `RBLN_DUMMY_DEVICE` (there is no NPU to probe for the SoC). The compile target is `RBLN_TARGET_SOC`, or the `npu=` `torch.compile` option when given.

## Tests

- `test/rbln/test_dummy_device.py` — dummy device contract: factories, copies, compile-only target resolution, multi-device shape, the compiled-graph execution guard, and the dummy-compile→real-NPU round trip.
- `test/cpp/core/RBLNDummyDeviceTest.cpp` — C++ vmem contract (11 subtests): alloc/transfer (sync + async), v2v-multi, owning-device resolution incl. interior pointers, borrow/try_borrow/acquire + return_borrowed, double/stale-free and out-of-bounds rejection, no-op layout/offload/synchronize.
- `test/models/test_vllm_llm.py` — vLLM native path (`VLLM_RBLN_USE_VLLM_MODEL=1`): `test_vllm_compile_only_no_npu_via_dummy` (build a servable model's `.rbln` with no NPU) and `test_vllm_dummy_compiled_runs_on_real_npu` (load those artifacts on a real NPU and generate, asserting no recompile). Skipped where `vllm-rbln` is not installed.

## Requires

rebel-compiler with the `ensure_valid_npu(npu)` contract + dummy-context support (pinned in `constraints-build-dev.txt`). vLLM's native attention lowers to a triton kernel, so the no-device→real vLLM path needs a rebel with the `rebel.triton` submodule (0.11.x line).

### DIFF
--- a/c10/rbln/RBLNFunctions.cpp
+++ b/c10/rbln/RBLNFunctions.cpp
@@ -456,6 +456,14 @@ void free_nothrow(void* data) noexcept {
 void set_device_layout_like(void* target_data, const void* ref_data) {
   RBLN_CHECK(target_data != nullptr, "set_device_layout_like: target is nullptr");
   RBLN_CHECK(ref_data != nullptr, "set_device_layout_like: ref is nullptr");
+  if (is_dummy_device()) {
+    // Host buffers have no device allocation layout to mirror; verify both
+    // pointers are live allocations on the same device, then no-op.
+    RBLN_CHECK(
+        dummy_lookup_device(target_data) == dummy_lookup_device(ref_data),
+        "set_device_layout_like: target and ref are on different dummy devices");
+    return;
+  }
   const auto target_vaddr = reinterpret_cast<uint64_t>(target_data);
   const auto ref_vaddr = reinterpret_cast<uint64_t>(ref_data);
   RBLN_LOG_DEBUG("set_device_layout_like: target={:#x} ref={:#x}", target_vaddr, ref_vaddr);
@@ -949,6 +957,10 @@ void reset_peak_memory_stats(const c10::Device& device) {
 }
 
 void set_file_offloading_enabled(bool enabled) {
+  if (is_dummy_device()) {
+    // File offloading is a runtime feature; no-op without a runtime (dummy mode).
+    return;
+  }
   RBLN_LOG_DEBUG("Calling rbln_set_file_offloading_enabled: enabled={}", enabled);
   RBLN_CHECK(
       !::rbln::rbln_set_file_offloading_enabled(enabled),
--- a/c10/rbln/RBLNFunctions.cpp
+++ b/c10/rbln/RBLNFunctions.cpp
@@ -5,10 +5,11 @@
 #include <c10/util/CallOnce.h>
 #include <rebel/runtime/memory_stats.h>
 
+#include <cstdint>
 #include <cstdlib>
 #include <cstring>
+#include <map>
 #include <mutex>
-#include <unordered_map>
 #include <vector>
 
 namespace c10::rbln {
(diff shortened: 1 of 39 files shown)


rebel-jonghewk:
@rebel-chanheo As far as I know, this doesn't require any compiler changes to support—is that right?

rebel-jongho (commented):
In this dummy NPU scenario, how is the number of available NPUs determined?

rebel-chanheo:
> @rebel-chanheo As far as I know, this doesn't require any compiler changes to support—is that right?

Yes, This is handled on the torch-rbln side; it doesn’t require compiler changes.

rebel-chanheo:
> In this dummy NPU scenario, how is the number of available NPUs determined?

In practice, the main dummy-mode use case is one logical device with an RSD/TP width, e.g. RBLN_NPUS_PER_DEVICE=N, for compile-only builds. RBLN_DEVICE_MAP can define multiple logical devices for topology parity/testing.

rebel-jongho (approved):
Looks good from the perspective of rebel-compiler runtime.

c10/rbln/RBLNFunctions.cpp
```diff
@@ -327,6 +343,11 @@ void free_nothrow(void* data) noexcept {
 void set_device_layout_like(void* target_data, const void* ref_data) {
   RBLN_CHECK(target_data != nullptr, "set_device_layout_like: target is nullptr");
   RBLN_CHECK(ref_data != nullptr, "set_device_layout_like: ref is nullptr");
+  if (is_dummy_device()) {
+    // Matching a device-physical allocation layout is meaningless without a
+    // real device; the host-backed buffers have no such layout. No-op.
+    return;
+  }
```
rebel-jonghewk: @rebel-chanheo Do you actually need to call this, or is this just an early return?

rebel-chanheo: Yes it's Needed. It's a public API (torch.rbln.set_device_layout_like) users can call in any mode; the body calls the device-only rbln_set_device_alloc_layout_like, which is meaningless on a dummy's host buffers, so it's a defined no-op there. Guarded at the C++ function (not Python) so direct C++ callers are covered — RBLNDummyDeviceTest.SetDeviceLayoutLikeIsNoOp relies on it.

rebel-jonghewk: >   meaningless on a dummy's host buffers

So just an extra overhead ? but not actual functionality issue?

rebel-chanheo: Not overhead — a correctness guard. `set_device_layout_like` configures target's device physical layout kind (NT/WT) to match ref, and the contract "ref must already have a physical view." Dummy buffers are host backed with no physical view, so `rbln_set_device_alloc_layout_like` returns failure and the `RBLN_CHECK` throws.

rebel-chanheo: @rebel-jonghewk 

Thank you for the feedback. Correction to my earlier comments — it's not a v-mem limitation.

- Dummy backs "device" allocations with v-memory, so `set_device_layout_like` isn't gated on a physical NPU: with a ref that has a physical view it runs to completion in dummy.
- What I saw earlier is just the first check `!ref->has_physical_view()` in `SetDeviceAllocLayoutLike`. In dummy the buffers I used had only a host user view, so it threw — but that's a general precondition (ref must be materialized), which a real NPU raises identically for an unmaterialized ref, Not dummy-specific.

→ So the `if (is_dummy_device()) return;` guard isn't a correctness requirement — it just turned that precondition throw into a silent no-op in dummy. Removing it so dummy stays consistent with real HW (raises the same precondition). I will fixing the test accordingly.

rebel-jonghewk (approved):
lgtm

PR #130, opened 2026-07-07
fix(device): gate runtime ops on runtime_available() so RBLN degrades like CUDA

rebel-chanheo:
## Problem

`torch-rbln` registers itself as the PrivateUse1 accelerator at `import` (via the `torch.backends` autoload entry point), so torch's device-agnostic APIs dispatch into it even when no device tensor is ever created. It links `librbln.so`, which loads the device runtime (driver) lazily. When that runtime is absent (compile / CPU-only / CI hosts) or torn down at interpreter shutdown, a raw `rbln_*` call dereferences a null handle → **uncatchable SEGFAULT** — unlike CUDA, where a missing driver returns an error code. vLLM's shutdown `torch.accelerator.empty_cache()` hits exactly this ([vllm-rbln#756](https://github.com/RBLN-SW/vllm-rbln/pull/756)).

## Solution

One nothrow liveness predicate — `runtime_available()` (runtime loaded, not shutting down, a device present) — is the single source of truth. It gates the runtime **entry points**, not every raw call, because most raw calls can't be reached with a dead runtime:

| op class | when the runtime is unavailable |
|---|---|
| best-effort entry points (`empty_cache` / `memory_stats` / `reset_*` / `set_file_offloading_enabled`) | **no-op** |
| `malloc` (allocation is the gateway) | **clean `RuntimeError`** via `require_runtime()` |
| `synchronize` | **throws when no device is usable** (no driver or no NPU; `torch.cuda` parity); no-op only on teardown |
| availability (`is_available` / `device_count` / `initialized` / `hasPrimaryContext`) | derive from it, **never segfault** |

`torch.accelerator.empty_cache()` then no-ops through torch's own `initialized()` guard, so **vLLM can drop its `empty_cache` monkeypatch**. A Python `atexit` hook flags shutdown so late frees leak instead of crashing at teardown.

**Runtime-loaded state comes from librbln.** The gate calls librbln's own `rbln_runtime_available()` ([rebel_compiler#12029](https://github.com/rebellions-sw/rebel_compiler/pull/12029)) directly, rather than torch-rbln probing the driver `.so` itself — authoritative (reflects librbln's real symbol-load state) and with no `dlopen`/soname coupling. Requires `rebel-compiler>=0.11.1.dev262+gfbe54616`.

**Why not gate every raw leaf?** The copy / borrow / layout leaves (`memcpy_*`, `borrow_host_ptr`, `set_device_layout_like`, …) all take a device address, which only exists after a successful `malloc` — and `malloc` is gated. So a dead runtime makes them unreachable; gating them would be dead code. The one runtime leaf reachable *without* an allocation is `set_file_offloading_enabled` (via `torch.rbln.offload()`), so it is gated directly.

**Root-cause fix for availability:** `device_count()` reaches a raw `rbln_get_device_count()` / (dummy) `rbln_register_device_id()` in `DeviceMappingManager`, so it would segfault before any Python-level guard. Enumeration is now gated on `rbln_runtime_available()` (before the dummy branch too) and degrades to 0 logical devices — fixing every enumeration entry point at the source (`torch.cuda.device_count() == 0` parity).

### Two failure modes, kept distinct

- **Runtime absent / torn down** → degrade gracefully (no-op / clean error / 0 devices).
- **Malformed `RBLN_DEVICE_MAP` / `RBLN_NPUS_PER_DEVICE`** → stay **loud** *when the runtime is present*: enumeration reaches config parse and throws (torch's `is_privateuse1_backend_available()` relies on this). With the runtime absent, enumeration degrades to 0 before parse — there is no device to configure anyway.

### Dummy mode

`RBLN_DUMMY_DEVICE` host-backs via the runtime, so it needs the runtime too — it is gated the **same** way everywhere (enumeration, allocation, free), proceeding when the runtime is present and degrading cleanly when not, not exempt.

## Testing

`test/rbln/test_runtime_unavailable.py` simulates a torn-down runtime via a process-wide shutdown flag (hardware-independent, runs in CI): best-effort no-op, `set_file_offloading_enabled` no-op, mandatory clean-throw, `is_available()` False, and dummy gating. A self-skipping test covers genuine runtime-absence (real + dummy) on compile-only lanes; `RBLNNoDeviceTest` covers the no-device `synchronize`/`malloc` throw contract. The full core suite (`test_device_mapping` / `test_no_device` / `test_dummy_device` / `test_file_offloading`) passes.

* Related to vllm-rbln#756, rebel_compiler#12029

### DIFF
--- a/c10/rbln/RBLNFunctions.cpp
+++ b/c10/rbln/RBLNFunctions.cpp
@@ -274,13 +274,13 @@ bool is_dummy_device() {
 // mandatory ops throw via require_runtime(), and nothing segfaults.
 
 // Is the RBLN device runtime (librbln-thunk.so) loaded and usable? Delegates to
-// librbln's own rbln_runtime_available(), which reports the actual thunk-load state
+// librbln's own rbln_runtime_available(), which reports the actual runtime-load state
 // (symbols resolved at librbln static-init) -- authoritative and safe to call when
-// the thunk is absent. Public so initialized()/hasPrimaryContext() can gate the
+// the runtime is absent. Public so initialized()/hasPrimaryContext() can gate the
 // throwing get_device_count() on it: with the runtime present a malformed
 // RBLN_DEVICE_MAP still throws (config parse runs); with it absent, enumeration
 // degrades to 0 before parse, so a missing runtime yields false without segfaulting.
-bool thunk_loadable() noexcept {
+bool runtime_loaded() noexcept {
   return rbln_runtime_available();
 }
 
@@ -288,11 +288,11 @@ namespace {
 
 std::atomic<bool> runtime_shutting_down_{false}; // set at teardown via a Python atexit hook
 
-// True when the runtime cannot be touched at all -- absent thunk or teardown -- as
+// True when the runtime cannot be touched at all -- absent runtime or teardown -- as
 // distinct from "no device present", which to_device_id() reports loudly. Lets a
 // teardown-safe op no-op without masking a genuine no-device error.
 bool runtime_torn_down() noexcept {
-  return runtime_shutting_down_.load(std::memory_order_relaxed) || !thunk_loadable();
+  return runtime_shutting_down_.load(std::memory_order_relaxed) || !runtime_loaded();
 }
 
 // Mandatory-op guard: a clean throw instead of a SEGFAULT (no-device is reported
@@ -301,7 +301,7 @@ void require_runtime(const char* op) {
   RBLN_CHECK(
       !runtime_shutting_down_.load(std::memory_order_relaxed), "Cannot {}: the RBLN runtime is shutting down.", op);
   RBLN_CHECK(
-      thunk_loadable(),
+      runtime_loaded(),
       "Cannot {}: the RBLN device runtime (librbln-thunk.so) is not loadable; install the RBLN SDK/runtime "
       "(required even in RBLN_DUMMY_DEVICE mode).",
       op);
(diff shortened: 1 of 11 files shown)


c10/rbln/RBLNFunctions.cpp
```diff
@@ -266,11 +268,83 @@ bool is_dummy_device() {
...
+// librbln.so uses (no ".so.N" version coupling); RTLD_LOCAL so a probe does not
+// promote thunk symbols globally. Public so initialized()/hasPrimaryContext() can
+// gate the throwing get_device_count() on it (config parsing is thunk-free, so a
+// malformed RBLN_DEVICE_MAP still throws while a missing runtime yields false).
+bool thunk_loadable() noexcept {
+  static const bool loadable = []() noexcept {
+    return ::dlopen("librbln-thunk.so", RTLD_LAZY | RTLD_LOCAL) != nullptr;
+  }();
+  return loadable;
+}
```
rebel-jonghewk: @rebel-chanheo I'm a bit hesitant about adding this layer into torch-rbln. Does it really need to know about this? Feels like it might belong elsewhere.

rebel-chanheo: @rebel-jonghewk I fully agree with your point. Probing `librbln-thunk.so` from torch-rbln is the wrong long-term layer boundary; torch-rbln should not need to know the thunk library name or librbln's lazy-loading details.

Would it be feasible to move this contract into the runtime layer instead? Ideally, public `rbln_*` APIs would never segfault when the thunk/driver is unavailable; they would return a normal error code, and torch-rbln could map that to `device_count() == 0`, `is_available() == False`, or a clean RuntimeError depending on the call. If changing all call sites is too large, a stable runtime-owned availability API such as `rbln_runtime_available()` / `rbln_get_runtime_status()` would also work.

For this PR, I'm fine treating the torch-rbln probe as a temporary crash barrier if we need an immediate fix, but I'd prefer not to bake `librbln-thunk.so` knowledge into torch-rbln permanently.

rebel-chanheo:
@rebel-jonghewk @rebel-eunji 

One additional release-timing concern: we are planning an RC release this weekend.

If this PR is included in that torch-rbln package, it should cover the problematic scenarios we have observed so far, especially missing-runtime env and shutdown-time cleanup paths. 
Could you help judge whether that short-term coverage is worth taking for the RC, or whether we should hold this PR and wait for the runtime-layer fix instead?

c10/rbln/RBLNFunctions.cpp
rebel-jonghewk: https://github.com/rebellions-sw/rebel_compiler/pull/12029

rebel-jonghewk (approved):
lgtm. Can you consider including this to RC?

rebel-chanheo:
> lgtm. Can you consider including this to RC?

Yes we will!

PR #136, opened 2026-07-10
fix(rbln): raise on empty-self index bounds; release fill_ borrow on overflow

rebel-chanheo:
## Summary of Changes

Two independent eager-mode correctness fixes surfaced by an audit:

- **index_select / index_copy on an empty `self`** — the index was bounds-checked only *after* a `numel()==0` early return, so an out-of-range index on an empty `self` returned uninitialized data (e.g. empty `self` `(0,3)` resized `out` to `(1,3)` and returned it) or was silently swallowed. Moved the check ahead of the early return so it raises like CPU. The check is gated on `source.numel() > 0` to match CPU, which validates index bounds only when there is data to copy — a valid index producing an empty output still succeeds, and an empty source never raises (even for a non-empty out-of-range index).
- **`fill_` / `arange` host-borrow leak on conversion overflow** — a checked cast can throw (e.g. `fill_(128)` into int8, or `arange` with a start/step out of an int8 out's range) *after* the borrow is acquired, skipping `return_borrowed` and leaking the borrow (failing the next free on that vaddr). A shared file-scope `BorrowGuard` releases the borrow on any exception in both `fill_scalar_rbln_` and `arange_start_out_rbln`; the normal paths still return explicitly and disarm.

Bundled because both are small, self-contained correctness fixes that turn silent-garbage / leak into correct behavior on previously-broken paths. The two fixes touch disjoint functions and are independent.

---

## Related Issues / Tickets

* Related to: eager-mode correctness audit (no tracking issue yet — please link one if required).

---

## Type of Change

- [x] `fix` — Bug fix

---

## Affected Modules

- [x] Ops/Kernels
- [x] Memory

---

## How to Test

1. `pytest test/rbln/test_cat_index_select_v2v.py test/rbln/test_index_copy_v2v.py test/rbln/test_native_fill_scalar.py`
2. Edge cases added: empty-`self` / empty-output out-of-range index raises on **both** RBLN and CPU (parity pinned); valid index with empty output does **not** raise; empty source with an out-of-range index does **not** raise (matches CPU); `fill_(128)` into int8 raises and the tensor stays usable afterward (borrow released).

---

## Checklist

* [x] Linting passes
* [x] All CI tests pass


aten/src/ATen/native/rbln/RBLNTensorFactories.cpp
rebel-daeyang, line 369: Same leak-on-throw as the `fill_` fix: `start.to<T>()`/`step.to<T>()` can throw on overflow, skipping `return_borrowed` (line 358) and leaking the borrow. Worth the same RAII guard.

test/rbln/test_cat_index_select_v2v.py
```diff
@@ -602,6 +602,31 @@ def test_index_select_oob_index_rejected(self):
         with pytest.raises(RuntimeError):
             torch.index_select(to_dev(a_cpu), 0, bad_idx)
 
+    def test_index_select_empty_self_oob_index_rejected(self):
+        """self empty on the indexed axis (extent 0): any index is out of range
+        and must raise, not resize out to (1,3) and return it uninitialized."""
+        empty_self = torch.empty((0, 3), dtype=torch.float32)
+        idx = torch.tensor([0], dtype=torch.long)
+        with pytest.raises(RuntimeError):
+            torch.index_select(to_dev(empty_self), 0, idx)
```
rebel-daeyang, line 613: Only assert the rbln path raises — CPU parity ("matches CPU") isn't actually checked. Consider `pytest.raises()` on the CPU tensor too, so parity is pinned in the test.

aten/src/ATen/native/rbln/RBLNTensorFactories.cpp
rebel-chanheo: Good catch — fixed in cd484e6. Hoisted the fill_ BorrowGuard to file scope and wrapped arange_start_out_rbln's borrow with it, so a throwing start.to<T>()/step.to<T>() can't leak. The normal paths still return explicitly and disarm.

test/rbln/test_cat_index_select_v2v.py
rebel-chanheo: Done in cd484e6 — added a pytest.raises on the CPU tensor too (both empty-self and empty-output cases), so CPU parity is pinned, not just the rbln path.

rebel-daeyang (approved):
LGTM

rebel-tskim (approved):

PR #137, opened 2026-07-10
fix(rbln): grow an undersized out= in the CPU fallback

rebel-chanheo:
## Summary of Changes

A pure `out=` slot is borrowed as a non-resizable `from_blob` view of the rbln v-memory. When the caller passes an **undersized** `out=`, the CPU kernel's resize-with-warning path tries to *grow* it and throws `"Trying to resize storage that is not resizable"`. Shrink already worked (metadata-only, synced back in Step 3); only grow failed.

Fix: catch that one error, release the offending pure-out borrow(s), swap in a fresh **zero-sized** resizable CPU tensor, and retry once via the reference kernel — Step 3's resize-writeback then grows the rbln `out` to match. Recreating the retry tensor zero-sized means the retry's grow-resize starts from an empty (numel==0) tensor, and the `"output was resized"` warning only fires when resizing a **non-empty** output — so the retry emits no additional warning.

Cost note: a correctly-sized `out=` never reaches the retry, but every pure-`out` borrow still walks the op schema once and snapshots the `Stack` (the snapshot is what lets the retry restore the consumed args). That is a constant, `out`-size-independent cost per pure-`out` fallback call — not a new cost specific to the undersized case. Its absolute magnitude is not benchmarked here.

## Related Issues / Tickets

* Related to: eager-mode correctness audit (no tracking issue yet — please link one if required).

## Type of Change

- [x] `fix` — Bug fix

## Affected Modules

- [x] Ops/Kernels

## How to Test

1. `pytest test/rbln/test_borrow_fast_path.py`
2. `torch.sin(torch.arange(8, device="rbln"), out=torch.empty(1, device="rbln"))` grows `out` to shape `(8,)` (was: raised `"not resizable"`), emitting the resize warning exactly once; oversized `out` still shrinks.

## Checklist

* [x] Linting passes
* [x] All CI tests pass

## Notes

The retry keys on the specific `"not resizable"` error message; any unrelated failure (dtype/device/etc.) propagates unchanged. The retry only runs when a pure-`out` is the sole write alias, so a mutable-`Tensor` argument is never re-mutated by the second kernel run.


aten/src/ATen/native/rbln/RBLNCPUFallback.cpp
```diff
@@ -472,19 +472,79 @@ void cpu_fallback_rbln(
...
+  auto run_cpu_kernel = [&]() {
+    auto fast_path_fn = CPUFastPathRegistry::instance().try_get(op.schema());
+    const bool fast_path_taken = fast_path_fn != nullptr && fast_path_fn(cpu_tensors, stack, arguments_begin);
+    if (!fast_path_taken) {
+      op.redispatchBoxed(c10::DispatchKeySet(cpu_dispatch_key), stack);
+    }
+  };
+
+  if (grow_retry_safe) {
+    const std::vector<c10::IValue> stack_snapshot = *stack;
```
rebel-tskim: It seems like we are allocating heap / copy-construct / etc per every correctly-sized borrows. I wonder if the overhead would be negligible.

rebel-chanheo: Good catch — switched the snapshot to a stack-inlined `c10::SmallVector<c10::IValue, 8>` (8 covers any pure-out op's arg count), so the heap alloc is gone for the common case.

The copy itself can't be dropped: `redispatchBoxed` consumes the stack, so we snapshot up front to be able to restore on the retry. Those are just IValue refcount bumps on a path only pure-out fallbacks reach (hot ops go through fast-path/device kernels), so the residual cost should be negligible.

rebel-tskim (approved):
LGTM!

PR #138, opened 2026-07-10
fix(rbln): harden SDPA cache lifecycle and reject dropout gradients

rebel-chanheo:
## Summary of Changes

RBLN SDPA caches the intermediate `attn_weights` in the forward (keyed by the output tensor's
`data_ptr()`) so the backward can reuse them instead of recomputing softmax. The cache write
was **unconditional**, but only the backward pops it — so under inference (`no_grad()`, where
no backward ever runs) every SDPA call stranded an `attn_weights` device tensor in a
process-lifetime dict. That is a memory leak on RBLN's primary (inference) path, and the
stranded prefill weights are large (`(batch, heads, seq, seq)`).

**Fix**
- Cache `attn_weights` **only when a backward will actually consume it** (`needs_backward =
  grad enabled and some input requires grad`). Inference under `no_grad()` no longer caches, so
  the inference leak is gone.
- The entry's lifetime is deliberately **not** tied to the output's Python wrapper. A
  grad-enabled forward whose backward never runs (e.g. `eval()` does not disable autograd)
  leaves its entry behind; repeated such forwards at distinct addresses can leave several.
  Each is bounded by output-address reuse: any later forward that reuses
  that `data_ptr` either **overwrites** it (a caching forward) or **drops** it (a non-caching
  forward — CPU fallback or fast-path inference — via `_discard_cached_attn_weights`), so a
  later backward can never pop stale weights left by a *different* forward. (Without the drop,
  a non-caching forward reusing the address left the stale entry, and its backward popped the
  wrong weights → shape error or silently wrong gradient; that path is now covered.) An earlier
  `weakref.finalize(output, …)`
  evictor was tried and reverted: the output's Python wrapper can be collected while the
  autograd graph still holds the tensor for a **pending** backward (the PyObject is not
  preserved for a saved-for-backward tensor), so the finalizer fired early and dropped a
  still-needed entry — turning `loss = sdpa(q, k, v).sum(); loss.backward()` into a silent CPU
  recompute. (The finalize callback also strong-ref'd `attn_weights`, stranding it until the
  output died even after backward popped the dict.)
- `torch.rbln.memory.empty_cache()` intentionally **preserves** a live entry that a pending
  backward still needs (forward → `empty_cache()` → backward must not lose it). It is not a
  flush point for this cache.

## dropout + autograd is rejected

With `dropout_p > 0` under autograd the RBLN gradient would be **numerically wrong** (not
merely imprecise): the forward applies dropout to `attn_weights`, but the backward recomputes
softmax from the *pre-dropout* weights and does not reproduce the forward dropout mask /
`1/(1-p)` scaling (`philox_seed`/`philox_offset` are not threaded into the backward) — the
gradient of a different function. Rather than return that silently, the forward now raises
`NotImplementedError`. The reject keys on `needs_backward` (which includes `attn_bias`), so a
graph required only through `attn_bias` is rejected too — closing the gap where the old
`can_use` gate checked only q/k/v. Inference dropout (`no_grad()` / `inference_mode()`) is unaffected — `eval()` does **not** disable autograd, so it is not a way around the reject.

Because the reject makes the upstream SDPA OpInfo dropout+grad samples raise, `test/filters.py`
wraps the SDPA OpInfo's `sample_inputs` to drop **only** the rejected samples — `dropout_p>0`
requested with `requires_grad` on an RBLN device (forward-only dropout and CPU samples are
kept) — so
`test_backward` / `test_variant_consistency_eager` / `test_noncontiguous_samples` still run
their `dropout_p==0` / supported-dtype samples (verified on NPU: 26 passed / 0 failed for
`-k scaled_dot_product`). Correctly threading the dropout mask into the backward (so grad is
supported, not rejected) is left as a follow-up.

## Type of Change

- [x] `fix` — Bug fix (inference memory leak)

## Affected Modules

- [x] Op/Dispatch
- [x] Memory

## How to Test

1. `pytest test/rbln/test_sdpa_decode_overflow.py` → green, including:
   - `test_inference_does_not_leak_attn_weights_cache` — SDPA under `no_grad` leaves the cache empty.
   - `test_grad_forward_then_backward_pops_cache` — forward caches, backward pops (output held in a live variable).
   - `test_grad_forward_anonymous_output_survives_until_backward` — an anonymous grad output's entry survives its wrapper's collection and is still present for backward (guards against re-tying the cache to the output wrapper).
   - `test_noncaching_forward_discards_stale_entry_at_reused_address` — a non-caching forward reusing an address drops the stale entry so a later backward can't pop another forward's weights.
   - `test_empty_cache_preserves_pending_backward_cache` — `empty_cache()` keeps a pending-backward entry.
   - `test_dropout_with_grad_rejected` / `test_dropout_with_attn_bias_grad_rejected` — dropout+autograd raises (incl. attn_bias-only grad); inference dropout still runs.
2. `pytest test/ops/test_ops.py -k scaled_dot_product` → green: the dropout+grad OpInfo sample is skipped for the affected consistency/backward tests, the rest pass.


torch_rbln/_internal/kernels/sdpa.py
```diff
@@ -694,11 +712,33 @@ def scaled_dot_product_fused_attention_overrideable_rbln(
...
+        or key.requires_grad
+        or value.requires_grad
+        or (attn_bias is not None and attn_bias.requires_grad)
+    )
+
+    # Dropout + autograd is rejected, not silently approximated: the backward recomputes softmax
+    # from the pre-dropout weights and can't reproduce the forward dropout mask / 1/(1-p) scaling
+    # (philox_seed/offset aren't threaded through), so the gradient would be of a different
+    # function. needs_backward includes attn_bias, so attn_bias-only grad is covered too.
+    if dropout_p > 0.0 and needs_backward:
```
rebel-tskim, line 726: This check seems to make R175 redundant. Let's remove R175 check if possible.

rebel-chanheo: Good catch — agreed, removed the `can_use_rbln_sdpa` gate (`dropout_p > 0 and requires_grad → "dropout with gradients"`).

rebel-huijong (approved):

rebel-scottlee (approved):
The existing SDPA cache leak is resolved for the primary inference paths using torch.no_grad() or torch.inference_mode().

A cache entry may still remain after a grad-enabled forward when backward is never executed. However, given torch-rbln’s inference-focused scope and the current low priority of backward performance, I do not consider this a blocker for this PR.

The remaining case can be tracked as a non-blocking follow-up, and the documentation should continue recommending torch.inference_mode() for inference workloads.

PR #141, opened 2026-07-10
fix(device): read TORCH_RBLN_EAGER_MALLOC live to prevent cross-test env leak

rebel-chanheo:
## Summary of Changes

`is_eager_malloc()` cached `TORCH_RBLN_EAGER_MALLOC` in a process-lifetime `static const bool`. The function-scoped `enable_eager_malloc` fixture toggles that env per-test, but the C++ cache latches the first-observed value for the whole process — so under pytest-xdist, the **first** test that allocates while the fixture is active pins the entire worker into eager mode, leaking it into every later, unrelated test on that worker.

This intermittently failed `test_runtime_hidden_d2h_is_cause_tagged` (`runtime_residency.total_count == 0` instead of `>= 1`: under eager, `y.copy_(x)` completes via device V2V with no hidden d2h). xdist shuffling the eager-latched worker per run is why it looked flaky.

- **Fix (`c10/rbln/RBLNFunctions.cpp`):** read the env live in `is_eager_malloc()` (drop the `static const bool`; compare with `strcmp(env, "1")`). Safe — only `malloc()` consults it and the free path (`rbln_free`) is mode-agnostic, so an alloc/free mode mismatch cannot occur; in production the env is fixed at startup so behavior is unchanged.
- **Make `test_runtime_hidden_d2h_is_cause_tagged` deterministic (not removed):** it now pins **lazy** allocation inside the test (`os.environ.pop("TORCH_RBLN_EAGER_MALLOC")` under `mock.patch.dict`) so the runtime always takes the `src_not_on_device` v2v hidden-d2h path, independent of the ambient/inherited env. The invariant it checks — the profiler counts the incident **and** attributes every hidden d2h to a named reason — is preserved.
- **C++ regression test (`test/cpp/core/RBLNFunctionsTest.cpp`):** `IsEagerMalloc` now actively asserts the live read: set → unset → set → `"0"` flips `is_eager_malloc()` on every call. A re-introduced process-lifetime cache would latch the first value and fail this, locking the fix against regression at the C++ level (no test-only wheel API needed).

## Related Issues / Tickets

* Related to # 

## Type of Change

- [x] `core` — device layer / C++ extension
- [x] `fix` — Bug fix

## Affected Modules

- [x] Device
- [x] C++ extension (`c10/rbln`)
- [x] Memory

## How to Test

1. C++ GTest `RBLNFunctionsTest.IsEagerMalloc`: asserts `is_eager_malloc()` reflects `set → unset → set → "0"` live (only exactly `"1"` is eager). Pre-fix, the `static` cache latched the first value and this would fail after the first toggle.
2. `pytest test/rbln/test_profiler.py` → green. `test_runtime_hidden_d2h_is_cause_tagged` now pins lazy allocation, so it is deterministic regardless of ambient `TORCH_RBLN_EAGER_MALLOC` (previously flaked when an xdist worker was latched into eager).
3. `TORCH_RBLN_EAGER_MALLOC=1` at process start → still eager; unset → lazy (production opt-in preserved).

## Notes

- The same latent pattern exists for `TORCH_RBLN_DEPLOY` / `is_deploy_mode()` (used by many tests). Not touched here (larger blast radius); the same live-read follow-up is recommended.
- The correlated `test_pinned_roundtrip_unaligned` failure (8194-byte unaligned sync H2V) is a **separate runtime issue** merely exposed by the eager leak; not addressed in this PR.

### DIFF
--- a/c10/rbln/RBLNFunctions.cpp
+++ b/c10/rbln/RBLNFunctions.cpp
@@ -340,7 +340,11 @@ c10::DeviceIndex get_torch_device_id(const void* data) {
 
 bool is_eager_malloc() {
   // Read live per call (not a process-lifetime static) so the value reflects the current
-  // environment. Safe: only malloc() consults it and free is mode-agnostic.
+  // environment. getenv is not thread-safe against a concurrent setenv/putenv, but this is
+  // safe here: only malloc() consults it, free is mode-agnostic (so an env change between an
+  // allocation and its free can't cause an alloc/free mismatch), and the env is not mutated
+  // concurrently with allocation (prod fixes it at startup; tests toggle it only at quiescent
+  // points, single-threaded per worker).
   const auto* env = std::getenv("TORCH_RBLN_EAGER_MALLOC");
   const bool eager_malloc = (env != nullptr) && (std::strcmp(env, "1") == 0);
   RBLN_LOG_DEBUG("eager_malloc={}", eager_malloc);
--- a/c10/rbln/RBLNFunctions.cpp
+++ b/c10/rbln/RBLNFunctions.cpp
@@ -339,11 +339,10 @@ c10::DeviceIndex get_torch_device_id(const void* data) {
 }
 
 bool is_eager_malloc() {
-  // Read live, not a process-lifetime static: enable_eager_malloc toggles this per-test
-  // (test/conftest.py); a static cache latches the first value into the whole xdist worker
-  // and leaks eager into later tests. Safe: only malloc() reads it, free is mode-agnostic.
+  // Read live per call (not a process-lifetime static) so the value reflects the current
+  // environment. Safe: only malloc() consults it and free is mode-agnostic.
   const auto* env = std::getenv("TORCH_RBLN_EAGER_MALLOC");
-  const bool eager_malloc = (env != nullptr) && (std::string(env) == "1");
+  const bool eager_malloc = (env != nullptr) && (std::strcmp(env, "1") == 0);
   RBLN_LOG_DEBUG("eager_malloc={}", eager_malloc);
   return eager_malloc;
 }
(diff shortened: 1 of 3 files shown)


rebel-chanheo:
Addressed the review (pushed d69d0f9):

- **set→unset→set regression** — `test_eager_malloc_env_is_live_not_process_latched`: toggles `TORCH_RBLN_EAGER_MALLOC` eager→lazy→eager and asserts the runtime tracks it live (eager → hidden-d2h `total_count == 0`, lazy → `>= 1`), proving the worker is no longer latched.
- **deterministic witness** — restored `test_runtime_hidden_d2h_is_cause_tagged` but now pins **lazy** allocation (delenv `EAGER_MALLOC`) so the `src_not_on_device` v2v hidden d2h reliably fires; it then asserts the incident is counted **and** fully cause-attributed. This is the real runtime integration coverage, no longer flaky under an eager-latched worker.

Both verified on-device (dev282).

c10/rbln/RBLNFunctions.cpp
```diff
@@ -339,10 +339,10 @@ ::rbln::MemoryInfo get_memory_info(const void* data) {
 }
 
 bool is_eager_malloc() {
-  static const bool eager_malloc = []() {
-    const auto* env = std::getenv("TORCH_RBLN_EAGER_MALLOC");
-    return (env != nullptr) ? (std::string(env) == "1") : false;
-  }();
+  // Read live per call (not a process-lifetime static) so the value reflects the current
+  // environment. Safe: only malloc() consults it and free is mode-agnostic.
```
rebel-tskim: eager_malloc is now not thread-safe (although it wouldn't matter on prod).
How about more verbose comment to prevent possible regression: `Safe: only malloc() consults it, free is mode-agnostic, env is not mutated concurrently with allocation (prod already fixes env on startup)`?

rebel-chanheo: Agreed. Reading the env live means each call does a getenv, which (unlike the previous cached static) isn't thread-safe against a concurrent setenv/putenv. It's safe as-is, but only under assumptions worth spelling out. so I'll expand the comment.

rebel-tskim (approved):

PR #150, opened 2026-07-13
build: source Rebel runtime headers from the rebel-compiler wheel

rebel-chanheo:
## Summary
The `rebel-compiler` wheel now ships the runtime API headers under `rebel/include`, making the vendored copy in `third_party/rebel_compiler/` redundant.

- **`cmake/FindRebel.cmake`**: resolve headers from `site-packages/rebel/include`, mirroring the existing `tvm/` lookup for `librbln.so`. Fail with a clear message if the installed wheel predates header shipping.
- Delete the vendored headers and drop their `third_party/` entry from the sdist include list.
- Bump the build pin (`constraints-build-dev.txt`) to `0.11.1.dev322+gbbd632a8.prod`, the first confirmed header-shipping build (`dev282`/`dev304` do not ship headers).

Headers and `librbln.so` now come from the same build, so they stay ABI-aligned by construction instead of via a manually-synced snapshot.

## Verification
- Confirmed the dev322 wheel (prod + non-prod) ships all 5 headers, byte-identical to the vendored copies.
- `find_package(Rebel)` resolves headers + library from a wheel-layout site-packages, and errors clearly when headers are absent.
- `dev322` is on the internal prod index (same lineage as the previous `dev304` pin), so the source build in CI resolves it.


rebel-huijong (approved):

rebel-clee (approved):
LGTM

rebel-tskim (approved):

PR #151, opened 2026-07-13
fix(profiler): move torch.profiler bridge registration to Python init and gate it by device type

rebel-ywkim:
## Summary of Changes

- Move the torch.profiler (kineto) bridge registration from C++ module init to Python init.
- Gate the registration by device type, enabling it only on supported devices.
- Log when rbln profiling is inactive so a skipped torch.profiler session is visible.

---

## Related Issues / Tickets

* Related to # https://github.com/rebellions-sw/rebel_compiler/pull/12079

---

## Type of Change

- [ ] `fix` — Bug fix

---

## Affected Modules

- [ ] Backend
- [ ] Device
- [ ] Ops/Kernels
- [ ] `torch.compile`
- [ ] Memory
- [ ] Build/Tools
- [ ] Docs

---

- On a supported device: run a model under `torch.profiler.profile(...)` and confirm rbln activities appear in the exported trace.
- On an unsupported device type: registration is skipped and a warning is logged at import; `import torch_rbln` and `torch.profiler` still run without error (no rbln activities in the trace).

---

---

## Checklist

---

## Notes

---

### DIFF
--- a/torch_rbln/__init__.py
+++ b/torch_rbln/__init__.py
@@ -194,14 +194,9 @@ def _create_process_group_rbln(dist_backend_opts, pg_options):
 
 def _initialize_kineto_profiler() -> None:
     """Register the rbln torch.profiler (kineto) bridge."""
-    try:
-        from rebel.device_info import get_npu_name
+    from torch_rbln._internal.device_arch_utils import is_atom_device
 
-        npu = (get_npu_name(0) or "").upper()
-    except Exception:
-        npu = ""
-
-    if npu.startswith("RBLN-CA"):  # ATOM
+    if is_atom_device():
         warnings.warn("rbln torch.profiler activities are not supported on ATOM and will be skipped", stacklevel=2)
         return
 
--- a/torch_rbln/__init__.py
+++ b/torch_rbln/__init__.py
@@ -195,6 +195,17 @@ def _create_process_group_rbln(dist_backend_opts, pg_options):
 def _initialize_kineto_profiler() -> None:
     """Register the rbln torch.profiler (kineto) bridge."""
     try:
+        from rebel.device_info import get_npu_name
+
+        npu = (get_npu_name(0) or "").upper()
+    except Exception:
+        npu = ""
+
+    if npu.startswith("RBLN-CA"):  # ATOM
+        warnings.warn("rbln torch.profiler activities are not supported on ATOM and will be skipped", stacklevel=2)
+        return
+
+    try:
         import torch_rbln._C
 
         torch_rbln._C._register_kineto_profiler()
--- a/torch_rbln/__init__.py
+++ b/torch_rbln/__init__.py
@@ -195,9 +195,10 @@ def _create_process_group_rbln(dist_backend_opts, pg_options):
 def _initialize_kineto_profiler() -> None:
     """Register the rbln torch.profiler (kineto) bridge."""
     from torch_rbln._internal.device_arch_utils import is_atom_device
+    from torch_rbln._internal.log_utils import rbln_log_info
 
     if is_atom_device():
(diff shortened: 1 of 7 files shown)


torch_rbln/__init__.py
```diff
@@ -198,6 +192,27 @@ def _create_process_group_rbln(dist_backend_opts, pg_options):
...
+def _initialize_kineto_profiler() -> None:
+    """Register the rbln torch.profiler (kineto) bridge."""
+    try:
+        from rebel.device_info import get_npu_name
+
+        npu = (get_npu_name(0) or "").upper()
+    except Exception:
+        npu = ""
+
+    if npu.startswith("RBLN-CA"):  # ATOM
```
rebel-jonghewk: FYI @rebel-chanheo, torch.profiler is only supported on REBEL

rebel-jonghewk (approved):

rebel-chanheo (commented):
Nice, clean change — moving registration to Python so the device gate can live where `get_npu_name` is easy to reach makes sense, and the fallback (register when the name is unknown) preserves the old always-on behavior. Two small, non-blocking thoughts inline. 🙏

torch_rbln/__init__.py
```diff
@@ -189,6 +192,27 @@ def _create_process_group_rbln(dist_backend_opts, pg_options):
...
+        warnings.warn("rbln torch.profiler activities are not supported on ATOM and will be skipped", stacklevel=2)
```
rebel-chanheo: Small UX thought: because `_initialize_kineto_profiler()` runs inside `torch_backends_entry_point()`, this `warnings.warn` fires on **every** `import torch_rbln` on ATOM — including for users who never touch `torch.profiler`. Would it be worth deferring the notice until profiling is actually attempted, or routing it through the quieter `RBLN_LOG_INFO` channel (as the C++ side already does) instead of a `UserWarning`?

One edge case that made me look twice: under `python -W error`, a `UserWarning` here would raise, propagate up to the `except Exception` in `torch_backends_entry_point`, and get re-raised — so `import torch_rbln` would fail outright on ATOM. Narrow, and the existing `Failed to register RBLN backend` warn has the same shape, so it's not new behavior — just flagging that import success ends up coupled to warning filters.

torch_rbln/__init__.py
```diff
@@ -189,6 +192,27 @@ def _create_process_group_rbln(dist_backend_opts, pg_options):
...
+def _initialize_kineto_profiler() -> None:
+    """Register the rbln torch.profiler (kineto) bridge."""
+    try:
+        from rebel.device_info import get_npu_name
+
+        npu = (get_npu_name(0) or "").upper()
+    except Exception:
+        npu = ""
+
+    if npu.startswith("RBLN-CA"):  # ATOM
```
rebel-chanheo: nit / single-source-of-truth: `test/utils.py` already encodes this exact arch mapping (`_arch_from_npu_name` / `is_atom_device`, `RBLN-CA*` -> atom). It lives in the test tree so it can't be imported from here, but this now adds a second copy of the `RBLN-CA` prefix + the `get_npu_name(0)` try/except in production code. Might be worth lifting a tiny helper into `torch_rbln/_internal/` and having both this and the tests share it, so ATOM detection has one home.

torch_rbln/__init__.py
rebel-ywkim: Thanks for the catching. I just replaced `warnings.warn` to `RBLN_LOG_INFO`.

As mentioned, the existing import-time `warnings.warn` calls that share this shape (`Failed to register RBLN backend`, the c10d-init failure, …) could get the same treatment, but I've left them as-is on this PR as I thought that would be out of scope for this PR. I can make a follow-up PR If we want to handle that case for the remaining functions as well.

rebel-chanheo (approved):
LGTM

PR #153, opened 2026-07-14
fix(device): track per-process context so memory ops match CUDA's initialized() contract

rebel-chanheo:
## Problem

`torch.accelerator.empty_cache()` aborted the caller with a `RuntimeError` at vLLM shutdown, crashing every DP `EngineCore`:

```
RuntimeError: rbln_empty_cache failed for rbln:0 (device may be busy or in a faulted state)
  cleanup_dist_env_and_memory() -> torch.accelerator.empty_cache() -> c10::rbln::empty_cache
```

Root cause: `RBLNAllocator::initialized()` reported *"a device/mapping exists"* (`runtime_available() && device_count>0`) instead of CUDA's *"this process has initialized the allocator."* torch guards the accelerator memory ops on this predicate (`_accelerator_isAllocatorInitialized`), so in the `EngineCore` parent — which has the runtime and a device mapping but **no live context** (its Worker child owns the device) — the guard passed and `empty_cache` dispatched into a contextless runtime, which fabricated a throwaway/dummy context and failed. The compile artifact was produced fine; the crash is at teardown, but it makes every DP process exit non-zero.

## Change

Track per-process context directly (CUDA parity with `device_allocator`): a per-logical-device bit is set on the first successful device `malloc` (the single allocation funnel; a two-word mask covers the full `DeviceIndex` range). It backs:

- `RBLNAllocator::initialized()` = `any_device_context_initialized() && runtime_available()`
- `RBLNHooksInterface::hasPrimaryContext(idx)` = `device_context_initialized(idx) && runtime_available()` — now **per-device** and shutdown-aware

The context flag is checked **first**, so a cleanup predicate on a no-context process never triggers device enumeration/registration as a side effect.

The **throwing** best-effort leaves (`empty_cache` / `reset_peak` / `reset_accumulated`) use a two-level gate, matching CUDA:

1. no allocator state anywhere in this process → no-op (never dispatches; covers the no-context parent **and** a malformed-config process),
2. allocator in use → `check_device_index` (an invalid index is a real error, throws),
3. this device unused → nothing to do,
4. otherwise run the runtime call and **propagate** a genuine failure (rc included) rather than swallowing it — an initialized allocator surfaces real errors.

The query leaves (`memory_stats` / `get_device_stats`) are left unchanged (they never throw); their pre-init empty result for `torch.accelerator.memory_stats()` comes from the same Python init guard via `initialized()`, so the existing `torch.rbln.memory_stats()` full-key contract is preserved.

This is *not* "CUDA `empty_cache` never raises": CUDA skips via its init guard **before** the allocator is used, but an initialized allocator propagates real errors — this matches that. Mandatory ops (malloc/free/copy/synchronize/…) are unchanged. RBLN device use after fork is unsupported (contexts don't survive fork); bad-fork detection is a follow-up.

## Tests

`test/rbln/test_runtime_unavailable.py` (subprocess-isolated, run on real NPU in CI):
- **no live context** → the ops never dispatch the runtime call (an `LD_PRELOAD` shim records non-invocation of empty_cache / reset_peak / reset_accumulated);
- **live-context failure** → propagates, per op, generic + direct C API, asserting the injected `rc=1` is reported;
- **malformed config** (real and dummy, `RBLN_NPUS_PER_DEVICE` and `RBLN_DEVICE_MAP`) → the predicate stays total and no-ops; the misconfig surfaces only at real device use;
- **lazy-malloc invariant** → a default-mode `torch.empty()` initializes the context and the ops then run.

`test/cpp/core/RBLNAllocatorTest.cpp`: `initialized()` / `hasPrimaryContext()` track allocation (per-device, invalid index, shutdown state); a tracker-bounds test (63/64/126 tracked, 127 invalid); the `*InvalidIndex` tests assert the throw; success tests allocate first so they exercise the real path independent of run order. Reset before/after effect is covered by `test_internal/test_memory_stats.py::test_reset_operations`.

The new subprocess tests fail on the pre-fix binary. Verified against the real runtime: gtest `RBLNAllocatorTest` 18/18 + `RBLNFunctionsTest` 32/33, Python suites green (incl. `test_memory_stats.py` in isolation).

Not included (cost): an end-to-end DP>1 vLLM `EngineCore` shutdown integration test — recommended as a required regression in downstream vLLM CI.

### DIFF
--- a/c10/rbln/RBLNFunctions.cpp
+++ b/c10/rbln/RBLNFunctions.cpp
@@ -861,10 +861,15 @@ void empty_cache(const c10::Device& device) {
   check_device_index(device_index);
   const auto device_id = to_device_id(device_index);
   RBLN_LOG_DEBUG("Calling rbln_empty_cache: device_id={}", device_id);
-  RBLN_CHECK(
-      !rbln_empty_cache(device_id),
-      "rbln_empty_cache failed for rbln:{} (device may be busy or in a faulted state)",
-      static_cast<int>(device_index));
+  // Best-effort: a failed cache flush must not abort the caller (CUDA parity — empty_cache
+  // never raises). A process with no live context makes the runtime fail here, and
+  // cleanup/teardown paths call this; warn instead. A real fault resurfaces at the next
+  // mandatory op (malloc/copy/execute, all hard-checked).
+  if (rbln_empty_cache(device_id)) {
+    RBLN_WARN_NOTHROW(
+        "rbln_empty_cache failed for rbln:{} (device may be busy or in a faulted state); skipping cache flush",
+        static_cast<int>(device_index));
+  }
 }
 
 std::map<std::string, uint64_t> memory_stats(const c10::Device& device) {
@@ -894,10 +899,11 @@ void reset_accumulated_memory_stats(const c10::Device& device) {
   check_device_index(device_index);
   const auto device_id = to_device_id(device_index);
   RBLN_LOG_DEBUG("Calling rbln_reset_accumulated_memory_stats: device_id={}", device_id);
-  RBLN_CHECK(
-      !rbln_reset_accumulated_memory_stats(device_id),
-      "rbln_reset_accumulated_memory_stats failed for rbln:{}",
-      static_cast<int>(device_index));
+  // Best-effort (see empty_cache): a failed reset carries no correctness meaning; warn, don't throw.
+  if (rbln_reset_accumulated_memory_stats(device_id)) {
+    RBLN_WARN_NOTHROW(
+        "rbln_reset_accumulated_memory_stats failed for rbln:{}; skipping reset", static_cast<int>(device_index));
+  }
 }
 
 void reset_peak_memory_stats(const c10::Device& device) {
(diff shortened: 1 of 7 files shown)


c10/rbln/RBLNFunctions.cpp
```diff
@@ -410,6 +411,38 @@ bool runtime_available() noexcept {
       (is_dummy_device() || get_device_count_nothrow() > 0);
 }
 
+// --- Per-process device-context tracking ------------------------------------
+// A per-logical-device bit set on the first successful device malloc, mirroring CUDA's
+// device_allocator populated on first use. Backs initialized()/hasPrimaryContext() and
+// gates the best-effort memory ops, so a process with the runtime + a mapping but no live
+// context (e.g. a vLLM EngineCore parent) reports uninitialized. Set-once, lock-free.
+// RBLN device use after fork is unsupported; bad-fork detection is not implemented yet
```
rebel-tskim, line 419: vllm-rbln relies on `spawn` for cases like TP, so bad-fork detection indeed seems unnecessary for now. Still, let's keep this noted.

rebel-tskim (approved):

PR #154, opened 2026-07-14
feat(rbln): torch.Stream / torch.Event support

rebel-chanheo:
## Summary of Changes

`torch.Stream` / `torch.Event` support for the RBLN (PrivateUse1) backend: the generic torch API, `torch.accelerator.*`, and a `torch.cuda`-style `torch.rbln.{Stream, Event, current_stream, default_stream, set_stream, stream}`. Consumer side of rebel-compiler [#12123](https://github.com/rebellions-sw/rebel_compiler/pull/12123), which expands the per-device default stream into a registry of independent seq-chains.

### Where it lives

Everything is in `RBLNGuardImpl`, so the three surfaces share one implementation. The only new pybind entry points are `_get_current_stream` / `_get_default_stream` / `_exchange_stream`, and they exist solely for the `torch.rbln` namespace.

| `RBLNGuardImpl` | C ABI |
|---|---|
| `getStream` / `exchangeStream` | `rbln_get_current_stream` / `rbln_set_current_stream` |
| `getDefaultStream` | none — `StreamId` 0 |
| `getNewStream` / `getStreamFromGlobalPool` | `rbln_stream_create`, pooled |
| `queryStream` / `synchronizeStream` | `rbln_stream_query` / `rbln_stream_synchronize` |
| `record` | `rbln_event_create` on first record, then `rbln_event_record` |
| `block` | `rbln_stream_wait_event`; cross-device falls back to `rbln_event_synchronize` |
| `queryEvent` / `synchronizeEvent` / `destroyEvent` | `rbln_event_query` / `rbln_event_synchronize` / `rbln_event_destroy` |

### Handles

A handle is `torch_device_id << 32 | local_id`. A `c10::Stream` carries `local_id` as its `StreamId` and the device in the `Stream` itself, so `StreamId` 0 is the default stream on both sides and a process that never creates a stream behaves as before. An event surfaces as its opaque 64-bit handle, which the runtime numbers from 1, so it never aliases the null "not created yet" sentinel.

### Decisions worth reviewing

| Decision | Why |
|---|---|
| `getNewStream` and `getStreamFromGlobalPool` both answer from a 32-deep per-device pool, one slot created per request | there is no destroy hook, so an unbounded create leaks and lengthens every `DrainAllStreams`. Same as CUDA; past the pool size, instances reuse a stream |
| `set_stream` selects the stream's device before exchanging the stream | matches `_cuda_setStream` and `torch.accelerator.set_stream`; otherwise allocations keep going to the old device |
| `getStream` answers the default stream only while the device has no context, and propagates failures after that | swallowing one would silently move work off the stream the caller selected. Creating a pool stream marks the context, so the shortcut cannot disagree with a stream a caller holds |
| `recordStream` stays a no-op | the runtime tags a freed block with the stream in flight at its free and folds foreign seqs into the reusing stream as cross-deps |

Rejected at the boundary: a stream whose `device_type` is not PrivateUse1 (`c10::Stream::unpack3` accepts any valid type), a `StreamId` wider than the handle's 32-bit field, a re-record onto another device's stream (as CUDA and XPU do), and an event handle of 0. `query` and `record` short-circuit during interpreter shutdown, like their synchronize counterparts.

### Not supported

| | Behavior |
|---|---|
| event timing | `elapsed_time` raises |
| stream priorities | accepted, ignored; `priority_range()` is `(0, 0)` |
| `Stream.native_handle` | unimplemented — raises on torch 2.11, absent from 2.10 |
| cross-device `wait_event` | degrades to a host-side synchronize, so it orders only the calling thread's later submissions |
| interprocess events, memory IPC | needs runtime support; see rebellions-sw/torch-rbln-internal#55 |

### DIFF
--- a/c10/rbln/RBLNFunctions.cpp
+++ b/c10/rbln/RBLNFunctions.cpp
@@ -757,10 +757,21 @@ c10::DeviceIndex resolve_device_index(c10::DeviceIndex device_index) {
   return device_index < 0 ? get_device_index() : device_index;
 }
 
-// Fixed per-device pool backing getStreamFromGlobalPool. RBLN has no stream
-// priorities, so one pool suffices; streams are created lazily and live for the
-// process (matching CUDA's global pool).
-constexpr int kStreamPoolSize = 32;
+// Fixed per-device pool behind every stream torch hands out. RBLN has no stream
+// priorities, so one pool suffices. Torch has no destroy hook for a stream, so
+// creating one per request would leak it and lengthen every runtime drain-all;
+// CUDA answers getNewStream from its pool for the same reason.
+constexpr size_t kStreamPoolSize = 32;
+
+c10::StreamId create_stream_local_id(c10::DeviceIndex device_index) {
+  const auto torch_device_id = static_cast<uint32_t>(to_device_id(device_index));
+  uint64_t handle = 0;
+  RBLN_CHECK(
+      !::rbln::rbln_stream_create(torch_device_id, &handle),
+      "rbln_stream_create failed for rbln:{}",
+      static_cast<int>(device_index));
+  return static_cast<c10::StreamId>(static_cast<uint32_t>(handle & 0xFFFFFFFFULL));
+}
 
 } // namespace
 
@@ -783,18 +794,6 @@ c10::Stream get_default_stream(c10::DeviceIndex device_index) {
   return c10::Stream(c10::Stream::DEFAULT, c10::Device(c10::kPrivateUse1, device_index));
 }
 
-c10::Stream new_stream(c10::DeviceIndex device_index) {
-  device_index = resolve_device_index(device_index);
-  check_device_index(device_index);
-  const auto torch_device_id = static_cast<uint32_t>(to_device_id(device_index));
-  uint64_t handle = 0;
-  RBLN_CHECK(
-      !::rbln::rbln_stream_create(torch_device_id, &handle),
-      "rbln_stream_create failed for rbln:{}",
-      static_cast<int>(device_index));
-  return stream_from_handle(device_index, handle);
-}
-
 c10::Stream get_stream_from_pool(c10::DeviceIndex device_index) {
   device_index = resolve_device_index(device_index);
   check_device_index(device_index);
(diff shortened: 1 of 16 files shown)


rebel-huijong (approved):
I'll update the generator side, referring to how create stream naturally acquires the device by going through to_device_id. Thank you.

```python
with stream(stream1): 
    big_work()
ev = stream1.record_event()

stream0.wait_event(ev)
```
Regarding the cross-device wait event: my understanding is that in CUDA, this code just enqueues s1's event wait on s0 and returns immediately, whereas in our case we implement it as a host sync, which makes it blocking. Is that correct?

rebel-chanheo:
> in our case we implement it as a host sync, which makes it blocking. Is that correct?

Correct. Same-device works because the event's seq and the target stream live in the same `Context`.
Cross-device has nothing to push. The seq belongs to the source device's `Context`, and every dep primitive is context-scoped, so the destination stream cannot reference a seq that is not its own. With no dep to enqueue, the only way left to honor the ordering is to wait on the host until that seq completes.

I'd expect this to be supported natively once the runtime grows cross-device deps.

torch_rbln/device/streams.py
```diff
@@ -0,0 +1,164 @@
...
+    """Set the current stream (no-op if ``None``). Prefer the :func:`stream` manager.
+
+    Also selects the stream's device, matching :func:`torch.cuda.set_stream`.
+    """
+    if stream is None:
+        return
+    torch_rbln._C._exchange_stream(stream.stream_id, stream.device_index, stream.device_type)
+
+
+class StreamContext:
```
rebel-jonghewk, line 130: Is `StreamContext` necessary? May be it can be simplified to this. What do you think? 

```
def stream(stream: Optional[Stream]) -> AbstractContextManager:
    return contextlib.nullcontext() if stream is None else stream
```

c10/rbln/RBLNFunctions.cpp
```diff
@@ -735,6 +735,189 @@ void synchronize(c10::DeviceIndex device_index) {
...
+void event_record(uint64_t event, c10::Stream stream) {
+  if (runtime_shutting_down_.load(std::memory_order_relaxed)) {
+    return;
+  }
+  RBLN_CHECK(
+      !::rbln::rbln_event_record(event, stream_to_handle(stream)), "rbln_event_record failed (event={:#x})", event);
+}
+
+void event_block(c10::Stream stream, uint64_t event) {
+  const auto stream_handle = stream_to_handle(stream);
```
rebel-jonghewk, line 958: I can see other having this guard
```
if (runtime_shutting_down_.load(std::memory_order_relaxed)) {
```

Is this not guarded on purpose?

torch_rbln/device/streams.py
rebel-chanheo: Mostly torch.cuda parity: 2.11 still has torch.cuda.stream() return a StreamContext even though the generic torch.Stream is a context manager itself, and this module exists to mirror that surface.

The other reason is where the enter/exit state lives — torch.Stream keeps it on the stream object, so two threads entering the same stream restore each other's previous stream, while StreamContext holds it per with.

For single-threaded use your version is equivalent and shorter, so if you'd rather trim it, Let me know and I'll simplify it.

c10/rbln/RBLNFunctions.cpp
rebel-chanheo: Not on purpose, just missed. Added in `3689e31`.

rebel-jonghewk (approved):
lgtm

PR #156, opened 2026-07-15
test(rbln): make test_custom_kernel user-level as custom ops move to rebel_compiler

rebel-chanheo:
## Summary

The `rbln_custom_ops` attention ops are being single-sourced into `rebel_compiler`, which owns the op definition, the converter, and the kernel math. torch-rbln owns only the PrivateUse1 eager dispatch path (`register_custom_ops.py`: namespace call → PrivateUse1 impl → compile → run). This test previously defined the ops itself and compared the real RBLN kernel against its own CPU body as a numerical golden — which no longer holds once torch-rbln does not own the definition.

## Change

Migrate `test_custom_kernel.py` to a user-level wiring / contract suite that exercises only what torch-rbln owns:

- Drop the local `@torch.library.custom_op` / `register_fake` definitions; `import rebel` single-sources registration.
- Keep the `torch.ops.rbln_custom_ops.*` calls — they still exercise the PrivateUse1 eager path (dispatch → compile → real kernel).
- Replace the tight numerical golden with contract checks: shape / dtype / device, declared in-place mutation, and a finite / non-degenerate sanity check.

Numerical correctness now lives in `rebel_compiler`, which owns the op definition, the converter, and the kernel.

## Scope

- `paged_attn_*` / `paged_causal_attn_*`: run against the current compiler (a missing op is a real regression and fails, by design).
- `flash_attention_naive_{prefill,decode}`: explicitly skipped until those ops are single-sourced in `rebel_compiler`; a follow-up drops the skip to enable them.

## Test

- Lint (lintrunner): clean.
- Full run needs an RBLN device; the suite runs under `@pytest.mark.test_set_ci`.

### DIFF
--- a/test/rbln/test_custom_kernel.py
+++ b/test/rbln/test_custom_kernel.py
@@ -1,667 +1,52 @@
 # Owner(s): ["module: PrivateUse1"]
 
-"""
-Test suite for RBLN custom kernel implementations.
+"""User-level wiring/contract tests for the rbln_custom_ops attention kernels.
 
-This test suite covers the following custom kernels:
-1. rbln_custom_ops::paged_attn_prefill
-2. rbln_custom_ops::paged_attn_decode
-3. rbln_custom_ops::paged_causal_attn_prefill
-4. rbln_custom_ops::paged_causal_attn_decode
-5. rbln_custom_ops::flash_attention_naive_prefill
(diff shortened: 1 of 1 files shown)


rebel-eunji (approved):
lgtm

PR #164, opened 2026-07-20
test(models): trim release model-test grid for LLMs (large + small)

rebel-chanheo:
## Why

The release model-test run (~2.6 h REBEL / ~3.6 h ATOM) is dominated by `test_transformers.py`, which swept a full `dtype × attn × batch × seq` grid on every model. Most of that grid re-runs the same torch-rbln code paths — model size and shape/dtype/attn interactions that op-level and CI tests already cover point-by-point — so it burns wall-clock without adding coverage.

## What

Two trims to the **release** transformers grid; the CI set is untouched.

**Large models (Llama-3B, EXAONE-2.4B)** — same architecture as the small models but 3–5× slower to compile. Drop the full grid; keep the representative point `(batch=2, seq=1024)` across the full `attn × dtype` matrix, plus a single largest-shape tiling smoke (`test_large_model_tiling`, `batch=4, seq=1024, fp16, sdpa`).

**Small models (Llama-1B, Qwen-1.5B)** — replace the full 4-way cross product (36 cases/model) with a **strength-2 (pairwise) covering array** (22 cases/model). It still guarantees:
- every `(batch, seq)` shape compiles (9 shapes — each a distinct tiling artifact);
- every dtype and every attn runs (1-way);
- every `attn×shape`, `dtype×shape`, `attn×dtype` pair appears (2-way);
- a full `dtype×attn` corner at `(2,1024)` (the CI point) and `(4,1024)` (max shape), where numeric-overflow / tiling issues concentrate.

Only the 3-/4-way interaction points are dropped.

## Coverage

- **CI set unchanged (16 → 16)** — only release-only cases are removed.
- Every torch-rbln code path still runs at least once. What's dropped is redundant re-runs: the large models duplicating the small-model architecture, and the small models' non-pairwise grid points.

## Numbers

| | before | after |
|---|---|---|
| release transformers cases | 144 | 54 |
| — large-model cases | 72 | 10 |
| — small-model cases | 72 | 44 |
| CI cases | 16 | 16 |

Case counts are exact; wall-clock to be confirmed on the next release run. Assuming ~fixed per-case compile cost, the model step should drop roughly ~30% beyond the large-model trim.

### DIFF
--- a/test/models/test_transformers.py
+++ b/test/models/test_transformers.py
@@ -169,6 +169,53 @@ class TestCausalLMBase(TestCase):
         torch.testing.assert_close(rbln_logits, cpu_logits, atol=self.LOGIT_ATOL, rtol=0.0)
 
 
+def _small_model_cover_array():
+    """Strength-2 (pairwise) covering array over dtype x attn x (batch, seq) for the small models.
+
+    The full 4-way cross product (2 dtype x 2 attn x 9 shapes = 36 cases/model) re-runs paths the
+    op-level and CI tests already cover point-by-point. This replaces it with a covering array of
+    22 cases/model that still guarantees:
+
+      * every (batch, seq) shape is compiled at least once (9 shapes -- each is a distinct
+        compiled artifact / tiling),
+      * every dtype and every attn runs (1-way),
+      * every attn x shape, dtype x shape, and attn x dtype pair appears (2-way),
+      * a full dtype x attn corner at the two largest shapes -- the CI point (2, 1024) and the
+        max shape (4, 1024), where numeric overflow / tiling issues concentrate.
+
+    Only the 3-/4-way interactions are dropped. Non-corner shapes get one row per dtype, rotating
+    the dtype<->attn pairing across shapes so all attn x dtype combinations are spread over the
+    small shapes too. The (2, 1024) rows carry the test_set_ci marker, keeping the CI set unchanged.
+    """
+    shapes = [(bs, sl) for bs in TestCausalLMBase.batch_sizes for sl in TestCausalLMBase.seq_lens]
+    attns = TestCausalLMBase.attn_implementations
+    full_corner_shapes = {(2, 1024), (4, 1024)}
+    n_dtype, n_attn = len(SUPPORTED_DTYPES), len(attns)
+
+    array = []
+    rotate = 0
+    for bs, sl in shapes:
+        if (bs, sl) in full_corner_shapes:
+            combos = [(dtype, attn) for dtype in SUPPORTED_DTYPES for attn in attns]
+        else:
+            combos = [
+                (SUPPORTED_DTYPES[(rotate + j) % n_dtype], attns[j % n_attn]) for j in range(max(n_dtype, n_attn))
+            ]
+            rotate += 1
+        is_ci = (bs, sl) == (2, 1024)
+        for dtype, attn in combos:
+            # Omit dtype from the subtest name: instantiate_device_type_tests appends it,
+            # which keeps the ids unique (e.g. ..._eager_b2_s1024_privateuse1_float16).
+            array.append(
+                subtest(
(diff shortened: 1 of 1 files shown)


rebel-chanheo:
@rebel-clee @rebel-huijong @rebel-tskim 

Purpose: Cut the release model-test runtime by dropping the redundant full test grid on the two large models (Llama-3B, EXAONE-2.4B) — same code paths as the small models — while keeping CI and overall coverage unchanged.

Reran the full release model suite on this branch — the model-test step dropped from ~3.6 h to ~2.5 h (a ~30% cut), and the two new tests (test_llama_3b, test_large_model_tiling) pass on real hardware.

The saving lands exactly where the measurement pointed: the two large models (Llama-3B, EXAONE-2.4B), which were ~47% of the model-test step, go from the full 36-combo batch × seq × attn × dtype grid down to the representative point + one largest-shape tiling smoke. Small models (Llama-1B, Qwen-1.5B), optimum, and vllm are untouched, and the CI set is unchanged (16 → 16).

test/models/test_transformers.py
```diff
@@ -169,17 +169,71 @@ def _assert_logits_match_fp32(self, model_id, config_kwargs, batch_size, seq_len
...
-    # CI runs only a representative subset (batch_size=2, seq_len=1024);
-    # release tests cover all batch_size x seq_len combinations.
-    ci_tests_batch_seq = [
-        subtest((bs, sl), decorators=[pytest.mark.test_set_ci] if (bs, sl) == (2, 1024) else [])
-        for bs in TestCausalLMBase.batch_sizes
-        for sl in TestCausalLMBase.seq_lens
-    ]
+    # Pin the EXAONE HF revision (shared by test_exaone and test_large_model_tiling) so
+    # model-hub updates can't shift the comparison.
+    _EXAONE_REVISION = "e949c91dec92095908d34e6b560af77dd0c993f8"
```
rebel-clee: `TestCausalLMPerf.test_exaone` still hardcodes the EXAONE revision, so it can drift from `_EXAONE_REVISION` the next time the pin changes. Would it make sense to hoist the constant to `TestCausalLMBase` so `TestCausalLMPerf` can share it?

rebel-clee:
Somewhat outside the scope of this PR, but worth raising while we're here:

The CI test set currently includes both `meta-llama/Llama-3.2-1B-Instruct` and `meta-llama/Llama-3.2-3B-Instruct` at the same representative point (batch=2, seq=1024), even though the two differ only in per-layer geometry under `num_hidden_layers = 1`.

Could we run one model per family in the CI tests, with the 3B model taking the Llama slot? Failures tend to surface first in the larger geometry, and the full 1B grid still runs in the release tests.

Happy to defer this to a follow-up.

test/models/test_transformers.py
rebel-chanheo: Good catch — hoisted _EXAONE_REVISION up to TestCausalLMBase so TestCausalLMPerf.test_exaone shares the same pin instead of hardcoding it

rebel-chanheo:
> Somewhat outside the scope of this PR, but worth raising while we're here:
> 
> The CI test set currently includes both meta-llama/Llama-3.2-1B-Instruct and meta-llama/Llama-3.2-3B-Instruct at the same representative point (batch=2, seq=1024), even though the two differ only in per-layer geometry under num_hidden_layers = 1.
> 
> Could we run one model per family in the CI tests, with the 3B model taking the Llama slot? Failures tend to surface first in the larger geometry, and the full 1B grid still runs in the release tests.
> 
> Happy to defer this to a follow-up.


Agreed, folded it in. CI now runs Llama-3B for the Llama family (failures surface first in the larger geometry); Llama-1B keeps its full release coverage (the pairwise covering array). CI set 16 → 12.

rebel-clee (approved):
LGTM

PR #168, opened 2026-07-21
fix(profiler): register the kineto bridge unconditionally so import doesn't seal RBLN_DEVICES

rebel-chanheo:
## Summary

Register the kineto profiler bridge unconditionally at import instead of gating it on `is_atom_device()`. The import-time arch query (`get_npu_name`) sealed `RBLN_DEVICES`, breaking vLLM data-parallel workers that remap it after import. ATOM profiling is now gated by the runtime, so nothing is lost.

## Problem

Reported on `vllm_rbln_exec.parity_runner --dp 4` with `RBLN_DEVICES=4,5,6,7`:

```
File ".../vllm_rbln/v1/worker/rbln_worker.py", line 195, in determine_available_memory
    device_name = current_platform.get_device_name().lower()
File ".../vllm_rbln/platform.py", line 116, in get_device_name
    rebel.get_npu_name(device_id)
RuntimeError: RBLN_DEVICES environment variable changed at runtime (Sealed)
```

Bisected to torch-rbln: **0.3.0rc0 OK, 0.3.0rc2 fails**. The rbln runtime seals `RBLN_DEVICES` on its first device-resolving call and raises if it later changes. A DP worker inherits a partition-wide `RBLN_DEVICES` and remaps it per rank *after* import, so anything that resolves the device at import breaks it.

## Root cause

#151 (`ea528145`, "skip registration on ATOM (unsupported)") moved the kineto bridge registration to Python init (runs at `import`) and gated it on `is_atom_device()` → `rebel.get_npu_name(0)`. That import-time query seals `RBLN_DEVICES` at the inherited value before the worker can remap it. rc0 registered the bridge in C++ with no arch query.

## Fix — #151's intent is preserved, moved to the correct layer

At the time of #151, the runtime did not gate ATOM, so the "skip on ATOM" check lived in torch-rbln — but it ran at import and sealed `RBLN_DEVICES`.

Since then, **rebel-compiler #12079 ("disable kineto export on ATOM SoCs")** moved that gate into the runtime: `rbln_kineto_is_active()` now returns inactive on ATOM, **side-effect-free** (it reads the already-registered profiling context's arch — it does **not** query the NPU name or seal `RBLN_DEVICES`). The C++ profiler `configure()` already gates session creation on `rbln_kineto_is_active()`, so ATOM sessions are still skipped.

So this PR removes the import-time `is_atom_device()` gate and registers unconditionally. **#151's intent (no rbln kineto profiling on ATOM) is preserved by the runtime, not dropped**; the import-time seal is gone. There are **no C++ changes** — `configure()`'s existing `is_active` check plus #12079 do the gating.

Requires rebel-compiler with #12079 (the pinned `dev416` includes it; merged 2026-07-13).

## Testing

`test/rbln/test_import_rbln_devices_seal.py`: after `import torch_rbln`, and after a CPU-only `torch.profiler`, remapping `RBLN_DEVICES` and querying the NPU must not raise the seal error.

### DIFF
--- a/test/rbln/test_import_rbln_devices_seal.py
+++ b/test/rbln/test_import_rbln_devices_seal.py
@@ -1,0 +1,98 @@
+# Owner(s): ["module: PrivateUse1"]
+
+"""Importing torch_rbln must not seal ``RBLN_DEVICES``.
+
+The rbln runtime seals ``RBLN_DEVICES`` on its first device-resolving call (e.g. an
+NPU-name query) and then raises ``RBLN_DEVICES environment variable changed at runtime
+(Sealed)`` if the value later changes. A vLLM data-parallel worker inherits a
+partition-wide ``RBLN_DEVICES`` and remaps it per rank *after* import, so import must not
+query the device arch.
+
+torch-rbln #151 gated the kineto profiler on ``is_atom_device()`` at import, which called
+``get_npu_name(0)`` and sealed ``RBLN_DEVICES`` (0.3.0rc0 OK, rc2 failed). The bridge now
+registers unconditionally; ATOM is gated by the runtime (``rbln_kineto_is_active()``
+reports inactive on ATOM, rebel-compiler #12079), not by an import-time arch query.
+"""
+
+import os
+import subprocess
+import sys
+import textwrap
+
+import pytest
+from torch.testing._internal.common_utils import run_tests, TestCase
+
+
+_PROJECT_ROOT = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
+
+
+@pytest.mark.test_set_ci
+class TestImportDoesNotSeal(TestCase):
+    """After import (or a non-RBLN profiler), ``RBLN_DEVICES`` must still be remappable.
+
+    Each subprocess sets ``RBLN_DEVICES``, runs the scenario, remaps it, then makes the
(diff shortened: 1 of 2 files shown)


rebel-chanheo:
@rebel-ywkim  I dropped your ATOM-skip commit (ea528145) in this PR.

Reason: that check runs at import and calls `get_npu_name()`, which seals RBLN_DEVICES. vLLM --dp remaps RBLN_DEVICES per rank after import, so it then hits changed at runtime (Sealed)

The ATOM intent isn't lost — the runtime handles it now: rebel #12079 makes `rbln_kineto_is_active()` return inactive on ATOM, and configure() already gates on `is_active`, so ATOM sessions are still skipped. So I just register unconditionally and let the runtime gate ATOM.

rebel-huijong (approved):

rebel-ywkim (approved):

rebel-ywkim:
The purpose of https://github.com/RBLN-SW/torch-rbln/commit/ea528145fe89f1a6a1fec0363ec4ace44153095b was initially to explicitly block kineto register for ATOM on torch-rbln side as well with some logs, so that user can be aware of registration being skipped explicitly, rather than silently disabled in rebel-compiler.

However, preserving that explicit torch-rbln side block costs more than it's worth, so I think dropping it looks good. 
Thank you for catching this issue and the fix. 🙏

PR #174, opened 2026-07-26
feat(generator): initial generator support

rebel-huijong:
# Summary

Added initial generator support for the RBLN device.

This PR implements `set_state` and `get_state` for `RBLNGenerator`. Since generator-consuming operators that are currently supported through the CPU fallback path require a CPU generator, each RBLN generator now maintains an associated CPU generator for fallback execution. The CPU fallback logic was also extended to recognize optional generator arguments and replace an RBLN generator with its corresponding fallback CPU generator.

Additionally, this PR adds persistent per-device default generators and exposes the necessary generator and fork-safety APIs through the RBLN device module.

# What's implemented

* Added `OptionalGenerator` to `CpuFbArgKind`.
* Added handling for `TypeKind::OptionalType` → `TypeKind::GeneratorType` in `get_or_populate_schema_info`.
* Added `OptionalGenerator` IValue handling in `cpu_fallback_rbln`, using `RBLNGenerator::get_fallback_generator()`.
* Implemented `set_state` and `get_state` for `RBLNGenerator`.
* Added an associated CPU generator to `RBLNGenerator` for use in the CPU fallback path.
* Implemented persistent per-device default RBLN generators.
* Added `manual_seed` and `manual_seed_all` to the RBLN device module.
* Added `_is_in_bad_fork` to the RBLN device module.
* Fixed `RBLNGenerator::seed` to properly set random seed instead of returning current seed.

# Tests

`test/rbln/test_generator_reproducibility.py`

# Notes

* This PR refines the approach proposed in [this issue](https://github.com/rebellions-sw/torch-rbln-internal/issues/46#issuecomment-4451294841).
* The generator-related logic introduced in this PR should be revisited once operators such as `aten::uniform_` and `aten::normal_ `can be handled natively without falling back to the CPU path.

### DIFF
--- a/aten/src/ATen/native/rbln/RBLNCPUFallback.cpp
+++ b/aten/src/ATen/native/rbln/RBLNCPUFallback.cpp
@@ -370,6 +370,44 @@ void cpu_fallback_rbln(
   std::vector<uint64_t> borrow_ids(tensor_args.size(), 0);
   std::vector<at::Tensor> cpu_tensors(tensor_args.size());
 
+  // If anything below throws (e.g. dtype-mismatched or wrong-device out=),
+  // the borrows above must still be released — else the next free on
+  // a still-borrowed vaddr fails, and since c10::rbln::free throws from
+  // a ~TensorImpl (noexcept) deleter that escalates to std::terminate.
+  // The RAII guard releases any still-live borrow (updated=false) on
+  // destruction; Step 4 zeroes borrow_ids on the happy path, so the guard
+  // is then a no-op.
+  struct BorrowReleaseGuard {
+    std::vector<uint64_t>& borrow_ids;
+    std::vector<std::vector<uint64_t>>& tensorlist_borrow_ids;
+    ~BorrowReleaseGuard() {
+      for (auto& bid : borrow_ids) {
+        if (bid != 0) {
+          try {
+            c10::rbln::return_borrowed(bid, /*updated=*/false);
+          } catch (...) {
+            // Best-effort: dtor is noexcept by default. Swallow runtime
+            // rejections so we don't escalate one borrow-release failure into
+            // std::terminate; the original exception that triggered cleanup
+            // is more useful for diagnosis.
+          }
+          bid = 0;
+        }
+      }
+      for (auto& bids : tensorlist_borrow_ids) {
+        for (auto& bid : bids) {
+          if (bid != 0) {
+            try {
+              c10::rbln::return_borrowed(bid, /*updated=*/false);
+            } catch (...) {
+            }
+            bid = 0;
+          }
+        }
+      }
+    }
+  } _borrow_guard{borrow_ids, tensorlist_borrow_ids};
+
   for (size_t i = 0; i < tensor_args.size(); ++i) {
     // Skip the borrow fast path for write-alias outputs (`out=`): from_blob's
     // fixed-size storage can't be resized, so the CPU op's resize path (wrong-shape
@@ -483,43 +521,6 @@ void cpu_fallback_rbln(
   // the borrowed out buffer, and replaces the stack; on no-match it returns
   // false and we fall through to the boxed dispatcher.
   //
-  // If Step 2/3 throws (e.g. dtype-mismatched or wrong-device out=), the borrows
-  // above must still be released — else the next free on a still-borrowed vaddr
-  // fails, and since c10::rbln::free throws from a ~TensorImpl (noexcept) deleter
-  // that escalates to std::terminate. The RAII guard releases any still-live
-  // borrow (updated=false) on destruction; Step 4 zeroes borrow_ids on the happy
-  // path, so the guard is then a no-op.
-  struct BorrowReleaseGuard {
-    std::vector<uint64_t>& borrow_ids;
-    std::vector<std::vector<uint64_t>>& tensorlist_borrow_ids;
-    ~BorrowReleaseGuard() {
-      for (auto& bid : borrow_ids) {
-        if (bid != 0) {
-          try {
-            c10::rbln::return_borrowed(bid, /*updated=*/false);
-          } catch (...) {
-            // Best-effort: dtor is noexcept by default. Swallow runtime
-            // rejections so we don't escalate one borrow-release failure into
-            // std::terminate; the original exception that triggered cleanup
-            // is more useful for diagnosis.
-          }
-          bid = 0;
-        }
-      }
-      for (auto& bids : tensorlist_borrow_ids) {
-        for (auto& bid : bids) {
-          if (bid != 0) {
-            try {
-              c10::rbln::return_borrowed(bid, /*updated=*/false);
-            } catch (...) {
-            }
-            bid = 0;
-          }
-        }
-      }
-    }
-  } _borrow_guard{borrow_ids, tensorlist_borrow_ids};
-
   // A pure-out borrow wraps the rbln out in a non-resizable from_blob view, so an
   // undersized out= the CPU kernel tries to GROW throws "not resizable" (shrink is
   // metadata-only and already handled in Step 3). Catch that one error, swap the
--- a/aten/src/ATen/native/rbln/RBLNCPUFallback.cpp
+++ b/aten/src/ATen/native/rbln/RBLNCPUFallback.cpp
@@ -9,6 +9,7 @@
 #include <ATen/native/rbln/RBLNTensorUtils.h>
 
 #include <c10/rbln/RBLNFunctions.h>
+#include <c10/rbln/RBLNGenerator.h>
 #include <c10/util/SmallVector.h>
 
 #ifndef AT_PER_OPERATOR_HEADERS
@@ -45,6 +46,7 @@ enum class CpuFbArgKind : uint8_t {
   TensorList,
   OptionalTensorList,
   Device,
+  OptionalGenerator,
 };
 
 struct CpuFbSchemaInfo {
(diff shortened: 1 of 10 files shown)


aten/src/ATen/native/rbln/RBLNCPUFallback.cpp
```diff
@@ -343,6 +352,13 @@ void cpu_fallback_rbln(
         tgt_device = ivalue.toDevice();
         (*stack)[arguments_begin + idx] = c10::IValue(c10::Device(kCPU));
         break;
+      case CpuFbArgKind::OptionalGenerator: {
+        if (ivalue.isGenerator()) {
```
rebel-chanheo: When generator is None, should we replace it with the fallback CPU generator associated with the target device’s default RBLN generator? 
Leaving it as None makes the CPU kernel consume the global CPU RNG state, so torch.rbln.manual_seed() and the new per-device default generators do not affect ordinary RBLN random operations. Using CPU for generation is expected; the concern is which CPU generator owns the RNG state.

```
torch.rbln.manual_seed(1234)
x = torch.randint(0, 100, (16,), device="rbln")

torch.rbln.manual_seed(1234)
y = torch.randint(0, 100, (16,), device="rbln")

assert torch.equal(x.cpu(), y.cpu())  # Not currently guaranteed
```

aten/src/ATen/native/rbln/RBLNCPUFallback.cpp
```diff
@@ -343,6 +352,13 @@ void cpu_fallback_rbln(
...
+          auto g = static_cast<RBLNGeneratorImpl*>(ivalue.toGenerator().unsafeGetGeneratorImpl());
```
rebel-chanheo: Could we validate the generator’s device type before this cast? A Generator? argument may contain a CPU generator, and casting its implementation to RBLNGeneratorImpl results in undefined behavior and may crash. This case should raise a clear device-mismatch error instead.

```
cpu_gen = torch.Generator(device="cpu").manual_seed(1234)
torch.randint(0, 100, (16,), device="rbln", generator=cpu_gen)

Expected: a RuntimeError indicating that an RBLN generator is required.
```

c10/rbln/RBLNGenerator.cpp
```diff
@@ -27,27 +38,94 @@ uint64_t RBLNGeneratorImpl::current_seed() const {
...
 
-void RBLNGeneratorImpl::set_state(const c10::TensorImpl& new_state) {}
+void RBLNGeneratorImpl::set_state(const c10::TensorImpl& new_state) {
+  TORCH_CHECK(new_state.device().is_cpu(), "RBLN generator state must be a CPU tensor, but got ", new_state.device());
+
+  TORCH_CHECK(new_state.dtype() == at::kByte, "RBLN generator state must be a ByteTensor, but got ", new_state.dtype());
+
+  const auto* state_ptr = static_cast<const uint8_t*>(new_state.data());
+
+  std::memcpy(&seed_, state_ptr, sizeof(seed_));
```
rebel-chanheo: Could we validate the state layout and size before reading this header? Generator.set_state() can receive an arbitrary CPU ByteTensor, but these two memcpy calls unconditionally read 16 bytes. A shorter state therefore causes an out-of-bounds read, and the size calculation below may also underflow.

Additionally, seed_ and offset_ are modified before the fallback CPU state is validated, so a failed set_state() may leave the generator partially mutated.

```
g = torch.Generator(device="rbln")
g.set_state(torch.empty(0, dtype=torch.uint8))
```

This should raise a clear RuntimeError. Please validate contiguity and the minimum state size before any reads, and commit seed_/offset_ only after the fallback state has been accepted.

torch_rbln/random.py
```diff
@@ -0,0 +1,32 @@
+"""torch_rbln Python bindings for random number generation.
+
+This module provides functions for managing random number generators on RBLN
+devices and integrates with PyTorch's accelerator random-seeding interface.
+"""
+
+import torch_rbln._C
+
+
+__all__ = [
```
rebel-chanheo, line 14: Should this module also expose `get_rng_state()` and `set_rng_state()`? PyTorch’s custom-device RNG interface uses these APIs to save and restore the default generator state.

Without them, standard utilities such as torch.random.fork_rng() fail:

```
with torch.random.fork_rng(device_type="rbln"):
    torch.rand(10, device="rbln")
```

This currently raises an AttributeError because torch.rbln.get_rng_state is missing. These functions could be thin wrappers around get_default_generator(device_index).get_state() and .set_state().

c10/rbln/RBLNGenerator.cpp
```diff
@@ -1,17 +1,28 @@
...
+  cpu_generator_->set_current_seed(seed_);
+}
+
+at::Generator RBLNGeneratorImpl::get_fallback_generator() const {
+  return at::Generator(cpu_generator_);
+}
 
 void RBLNGeneratorImpl::set_current_seed(uint64_t seed) {
   seed_ = seed;
+  cpu_generator_->set_current_seed(seed);
```
rebel-chanheo, line 40: Could accesses to the associated fallback CPU generator be synchronized using its own mutex?

Python generator APIs lock the outer RBLN generator mutex, while fallback CPU random kernels lock the associated CPU generator’s mutex. Since these are different mutexes, state manipulation and sampling can access the same CPU RNG state concurrently.

```
from concurrent.futures import ThreadPoolExecutor

g = torch.Generator(device="rbln").manual_seed(1234)

with ThreadPoolExecutor(max_workers=2) as pool:
    pool.submit(lambda: [g.manual_seed(i) for i in range(1000)])
    pool.submit(
        lambda: [
            torch.rand(1024, device="rbln", generator=g)
            for _ in range(1000)
        ]
    )
```

Could we ensure that state APIs and fallback sampling synchronize access to the associated fallback CPU generator through the same mutex?

test/rbln/test_generator_reproducibility.py
```diff
@@ -0,0 +1,66 @@
...
+import pytest
+import torch
+from torch.testing._internal.common_device_type import instantiate_device_type_tests
+from torch.testing._internal.common_utils import run_tests, TestCase
+
+import torch_rbln  # noqa: F401  (registers the rbln backend)
+
+
+@pytest.mark.test_set_ci
+class TestGeneratorBasic(TestCase):
```
rebel-chanheo, line 57: Could we broaden the coverage beyond the explicit-generator happy paths? This PR also introduces default per-device generators, fallback-generator translation, and state parsing, but those paths and error cases are not currently exercised.

At minimum, it would be useful to cover:

- reproducibility after torch.rbln.manual_seed() without an explicit generator=
- isolation between the CPU default RNG state and RBLN RNG state
- rejection of a generator from another backend
- rejection of short or non-contiguous RNG states
- per-device generator independence when multiple RBLN devices are available
- state restoration through torch.random.fork_rng() if get_rng_state() and set_rng_state() are added

rebel-chanheo (changes requested):

aten/src/ATen/native/rbln/RBLNCPUFallback.cpp
rebel-huijong: You're right, when None is passed in, we should use the default RBLN generator (or more precisely, its CPU fallback generator). Will fix this right away.

aten/src/ATen/native/rbln/RBLNCPUFallback.cpp
rebel-huijong: I'll check the GPU error message for this part and handle it similarly.

c10/rbln/RBLNGenerator.cpp
rebel-huijong: I'll add the corresponding TORCH_CHECKs earlier in the function

torch_rbln/random.py
rebel-huijong: I'll add this as well.

c10/rbln/RBLNGenerator.cpp
rebel-huijong: I completely overlooked this one. I'll give it some thought and come back with a fix.

test/rbln/test_generator_reproducibility.py
rebel-huijong: I'll address all the comments above, and once the feature is complete, I'll add tests covering all of it.

aten/src/ATen/native/rbln/RBLNCPUFallback.cpp
```diff
@@ -429,6 +443,38 @@ void cpu_fallback_rbln(
...
+
+    c10::DeviceIndex tgt_index = tgt_device.has_value() ? tgt_device->index() : static_cast<c10::DeviceIndex>(-1);
+    if (tgt_index == -1) {
+      tgt_index = c10::rbln::get_device_index();
+    }
+
+    const auto& ivalue = arguments[generator_arg_idx.value()];
+    RBLNGeneratorImpl* arg_or_default_generator_ptr = nullptr;
+    if (ivalue.isGenerator()) {
+      arg_or_default_generator_ptr = check_generator<RBLNGeneratorImpl>(ivalue.toGenerator());
```
rebel-chanheo, line 497: At this point, RBLN tensor inputs may already have been borrowed above, but BorrowReleaseGuard is not constructed until after this block. If check_generator() or the device-index check throws, those borrows are never returned.

For example:

```
mean = torch.zeros(1024, device="rbln")
cpu_gen = torch.Generator(device="cpu")
torch.normal(mean, 1.0, generator=cpu_gen)

```
`normal.Tensor_float` is registered to this fallback, so mean is borrowed before the foreign generator is rejected. Releasing the last tensor reference afterward can then fail on the outstanding borrow and potentially terminate the process.

Could we validate the generator before acquiring any tensor or TensorList borrows, or install the cleanup guard before the first possible borrow? Please also add a tensor-input rejection test; the current randint test has no input tensor to borrow.

c10/rbln/RBLNGenerator.cpp
rebel-chanheo: Thanks for adding the layout and size checks. The partial-mutation case still remains: `seed_` is copied before `CPUGeneratorImpl::set_state()` validates the MT19937 payload.

```
g = torch.Generator(device="rbln").manual_seed(13)
before = g.get_state()
bad = torch.zeros_like(before)  # correct size/layout, invalid MT19937 state

with pytest.raises(RuntimeError, match="Invalid mt19937 state"):
    g.set_state(bad)

assert torch.equal(g.get_state(), before)  # currently fails
```

Could we parse the header into a temporary and assign seed_ only after the fallback state has been accepted? The failed-state test should include a correctly sized but semantically invalid payload.

aten/src/ATen/native/rbln/RBLNCPUFallback.cpp
rebel-huijong: I didn't have a good grasp of this part. Thank you for the explanation. Will fix this.

c10/rbln/RBLNGenerator.cpp
rebel-huijong: Ah, I missed this too. As you suggested, I'll fix it so there's no partial-mutation case.

rebel-huijong:
@rebel-chanheo 
Since the runtime API device query for the default generator is the problematic part, how about we set it aside for now in this PR and start by supporting only the explicit-generator-as-arg path? Just that would probably cover vLLM's dummy weight loading.

PR #175, opened 2026-07-28
feat(memory): add torch.rbln.release_offload_temp_storage()

rebel-jonghewk:
## What

Adds `torch.rbln.release_offload_temp_storage() -> int`, which removes the file-offloading temp
files and per-process directory this process created, returning the number of files removed.

## Why

`torch.rbln.offload()` writes into a per-process dir under `RBLN_OFFLOAD_DIR` (default
`$HOME/.cache/rbln_cache/offload`). The rebel runtime removes that dir on teardown, but the
teardown hangs off a Python `atexit` hook — the last thing to run. vLLM escalates to SIGKILL a few
seconds after asking a worker to stop, so a worker killed mid-shutdown never reaches it and leaves
the dir behind; one DP4 run left 4.7 GB. This gives a shutdown path a way to reclaim the space up
front instead of waiting for teardown.

## How

Thin wrapper over rebel's `rbln_release_offload_temp_storage`, bound as
`_C._release_offload_temp_storage`. Gated on `runtime_available()` like the offload toggle, so it
is a no-op when there is no runtime.

Contract: offloaded tensors must not be used afterwards — their vmemory entries survive but the
on-disk backing is gone. A later offload write allocates a fresh dir, so the manager stays usable.
Only this process's own dir is touched; dirs left by a killed run are still reclaimed lazily by the
sweep the runtime does when it creates the next offload dir.

## Tests

Three tests in `test_file_offloading.py`: exposed in `__all__` and on `torch.rbln`, idempotent, and
`offload()` still usable after a release. The mechanism itself is covered by rebel-compiler's
`vmem_file_offloading_test.cc`.

## Related

* rebel-compiler: rebellions-sw/rebel_compiler#12421 — adds `rbln_release_offload_temp_storage`
* vllm-rbln: RBLN-SW/vllm-rbln#836 — the consumer, calls this from `RBLNWorker.shutdown()`

Merge order is bottom-up: rebel-compiler first, then this PR once a rebel-compiler wheel carrying
the new symbol is pinned in `constraints-build-dev.txt`, then vllm-rbln. CI here fails until then —
`FindRebel` takes headers from the installed wheel, which does not declare the symbol yet.

### DIFF
--- a/test/rbln/test_file_offloading.py
+++ b/test/rbln/test_file_offloading.py
@@ -22,6 +22,8 @@ The flag and the depth counter are both process-local, so each pytest-xdist
 worker is isolated; the class does not need ``single_worker``.
 """
 
+from unittest.mock import patch
+
 import pytest
 import torch
 from torch.testing._internal.common_utils import run_tests, TestCase
@@ -107,6 +109,13 @@ class TestFileOffloading(TestCase):
         num_removed = torch.rbln.release_offload_temp_storage()
         self.assertIsInstance(num_removed, int)
         self.assertGreaterEqual(num_removed, 0)
+
+    def test_release_offload_temp_storage_returns_binding_count(self):
+        """The wrapper is a pass-through: the binding's count is returned unchanged."""
+        with patch.object(torch_rbln_C, "_release_offload_temp_storage", return_value=7) as mock_release:
+            self.assertEqual(torch.rbln.release_offload_temp_storage(), 7)
+
+        mock_release.assert_called_once_with()
 
     def test_release_offload_temp_storage_is_idempotent(self):
         """Releasing twice is safe; the second call has nothing left to remove."""
--- a/test/rbln/test_file_offloading.py
+++ b/test/rbln/test_file_offloading.py
@@ -94,6 +94,34 @@ class TestFileOffloading(TestCase):
 
         self.assertEqual(torch_rbln_memory._offload_depth, 0)
 
+    # ------------------------------------------------------------------
+    # Temp storage release
+    # ------------------------------------------------------------------
+    def test_release_offload_temp_storage_surface(self):
+        """``release_offload_temp_storage`` is exposed and reports a file count."""
+        from torch_rbln.memory import release_offload_temp_storage  # noqa: F401
+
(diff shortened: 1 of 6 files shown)


test/rbln/test_file_offloading.py
```diff
@@ -94,6 +94,34 @@ def test_offload_restores_depth_on_exception(self):
...
+        """``release_offload_temp_storage`` is exposed and reports a file count."""
+        from torch_rbln.memory import release_offload_temp_storage  # noqa: F401
+
+        self.assertIn("release_offload_temp_storage", torch_rbln_memory.__all__)
+        self.assertTrue(hasattr(torch.rbln, "release_offload_temp_storage"))
+
+        num_removed = torch.rbln.release_offload_temp_storage()
+        self.assertIsInstance(num_removed, int)
+        self.assertGreaterEqual(num_removed, 0)
+
```
rebel-chanheo, line 112: Non-blocking: Since this PR primarily adds an API bridge, could we mock `_C._release_offload_temp_storage()` to return a sentinel such as 7 and assert that the public wrapper returns it unchanged? The current non-negative integer assertion would also pass if the wrapper accidentally returned a constant 0.

test/rbln/test_file_offloading.py
```diff
@@ -94,6 +94,34 @@ def test_offload_restores_depth_on_exception(self):
...
+    def test_release_offload_temp_storage_surface(self):
+        """``release_offload_temp_storage`` is exposed and reports a file count."""
+        from torch_rbln.memory import release_offload_temp_storage  # noqa: F401
+
+        self.assertIn("release_offload_temp_storage", torch_rbln_memory.__all__)
+        self.assertTrue(hasattr(torch.rbln, "release_offload_temp_storage"))
+
+        num_removed = torch.rbln.release_offload_temp_storage()
+        self.assertIsInstance(num_removed, int)
+        self.assertGreaterEqual(num_removed, 0)
```
rebel-chanheo, line 111: Non-blocking: Could we also cover the unavailable-runtime contract in test_runtime_unavailable.py by setting the runtime-shutting-down flag and asserting that `torch.rbln.release_offload_temp_storage() `returns 0?

rebel-chanheo (approved):
LGTM with nits

PR #179, opened 2026-07-30
fix(torch.compile): skip tensor-value repr in Dynamo ID_MATCH guard build

rebel-jonghewk:
## Summary
Dynamo's guard build reprs weight `nn.Parameter`s to fill a cosmetic `type=` string. For an rbln weight-free (external-only) weight that reconstructs the whole tensor (D2H) on every repr, so `torch.compile` warm-up floods D2H and hangs. Scope the repr to return the *type* (not the value) only during guard build; user-facing `repr(tensor)` is unchanged.

## Problem
`device_tensor` + weight-free warm-up floods `[rbln_memcpy] copy_type=1` (D2H) and hangs:
```
build_guards → ID_MATCH → repr(weight) → get_cpu_copy_of_rbln_tensor
  → CopyVirtualToHost → ReconstructRegionFromExternalRefs (D2H)
```

## Root cause
`torch/_dynamo/guards.py` `id_match_unchecked` does `type_repr = repr(val)` — but the guard's real check is `id(val)`; `type={type_repr}` is verbose-only. `repr(tensor)` materializes the tensor (fp8/bf16 go through a full `.float()`), and an external-only weight reconstructs from device on each one. Fixed upstream (`repr(type(val))`) as of torch 2.13; 2.11 and 2.12 still `repr(val)`.

## Fix
`patch_dynamo_guard_repr()` in `_internal/monkey_patches.py`, wired into `apply_all_patches()`: wrap `id_match_unchecked` with a thread-local flag and have `Tensor.__repr__` return `repr(type(self))` only while it's set. Guard build never needs the value; the `id()` check and recompile triggers are untouched, and `repr(tensor)` outside guard build is unchanged. Gated on `torch.__version__ >= (2, 13)` so it's a no-op once the fix ships; `remove_all_patches()` restores.

Measured (DeepSeek-V3 fp8 weight-free): `copy_type=1` D2H **6400 → 32**, init engine **~170s → 116s**.

## Testing
`test/rbln/test_torch_compile_patch.py::TestDynamoGuardReprPatch` — idempotency, user repr unchanged, guard-scope returns type (no materialization), `remove_all_patches` restore, and an eager-backend `torch.compile` smoke asserting the guarded Parameter is never value-repr'd during guard build (output stays correct). Python lint clean.

Resolves rebellions-sw/torch-rbln-internal#45

### DIFF
--- a/torch_rbln/_internal/monkey_patches.py
+++ b/torch_rbln/_internal/monkey_patches.py
@@ -1,5 +1,6 @@
 """Monkey patches applied to PyTorch to enable RBLN functionality."""
 
+import threading
 import warnings
 
 from torch_rbln._internal.compile_cache import clear_rbln_compile_cache
@@ -12,6 +13,10 @@ _torch_dynamo_reset_patched: bool = False
 _rbln_backend_registered: bool = False
 _original_torch_compile = None
 _original_dynamo_reset = None
+_dynamo_guard_repr_patched: bool = False
+_original_id_match_unchecked = None
+_original_tensor_repr = None
+_guard_repr_tls = threading.local()
 
 
 def _is_backend_registered(backend_name: str) -> bool:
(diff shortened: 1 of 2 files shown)


torch_rbln/_internal/monkey_patches.py
```diff
@@ -141,9 +146,51 @@ def reset_wrapper(*args, **kwargs):
         _torch_dynamo_reset_patched = True
 
 
+def patch_dynamo_guard_repr() -> None:
+    """Keep Dynamo guard build from materializing tensor values via ``repr``.
+
+    Substitute value->type only during guard build; user ``repr(tensor)`` is
+    unchanged. Backport of the pytorch-main fix; remove once torch >= 2.12.
```
rebel-jongho: 이거 코드로 박아둘수는 없을까요?

rebel-jonghewk: Done (`d1cbd2a`) — hard-coded it in the patch rather than leaving it as a docstring note:

```python
import torch
# Fixed upstream in torch 2.13 (id_match_unchecked reprs the type, not the value)
if torch.__version__ >= (2, 13):
    return
```

`torch.__version__` is a `TorchVersion`, so the tuple comparison works, and the patch becomes a no-op on torch >= 2.13. (Verified against the release branches: 2.11 and 2.12 still `repr(val)`; 2.13 and main are fixed — so the gate is 2.13, not 2.12.)

rebel-jonghewk: I'll apply this.

rebel-huijong:
cc @rebel-cjlee

rebel-chanheo (approved):
LGTM with nits,

Given the urgency, I’m okay with landing this as a temporary workaround. As a follow-up, it would be better to replace the process-wide `Tensor.__repr__` monkey patch with a version-gated, guard-local backport of the upstream id_match_unchecked change. This would match the upstream behavior more closely and reduce global side effects. 

We can address that in a separate PR and remove the workaround once our minimum PyTorch version is 2.13+.

rebel-huijong (approved):

PR #191, opened 2026-08-06
feat(memory): expose per-chiplet memory stats and memory_summary()

rebel-jonghewk:
### Summary

`memory_stats()` aggregates across chiplets. An allocation fails on the heaviest chiplet of a device while the aggregate still looks free, so the numbers cannot explain the OOM. The allocator already tracks stats per chiplet — only the aggregation hid them. This adds `memory_stats_per_chiplet()` and `memory_summary()`.

A logical device can map to more than one physical NPU (`RBLN_NPUS_PER_DEVICE`, `RBLN_DEVICE_MAP`), so the breakdown is keyed on both axes, not on chiplet alone.

### Details

- **`memory_stats_per_chiplet()`** — Same keys as `memory_stats()`, each prefixed with `npu.<n>.chiplet.<c>.`. `npu.<n>` is the n-th NPU of *this* logical device, not a physical NPU id, so a 1:1 mapping yields `npu.0` only. It returns a flat dict rather than a new type: `c10::CachingDeviceAllocator::DeviceStats`, what `get_device_stats()` returns, is a fixed torch struct with no chiplet axis. Keeping the dict contract means no new pybind type and no change to `memory_stats()`.

- **`memory_summary()`** — The one `torch.cuda` counterpart still missing here. One row per (NPU, chiplet) plus a total, so an imbalance is visible at a glance. Peaks are tracked per chiplet, so the peak columns of the `total` row bound the joint peak from above rather than reporting it — `memory_stats()` carries the exact joint peak.

- **Scope is now stated in the output and the docstrings** — These are the caching allocator of the context *this process* holds on the device, the same scope `torch.cuda.memory_stats()` reports. Another process on the same NPU has its own context and its own numbers. A table titled `device=rbln:0` invites reading it as the NPU's occupancy, so the summary carries a scope line and the docstrings point at `rbln-smi` for a device-wide figure.

- **Gating** — Both go through the same gates as `memory_stats()`: `runtime_available()`, `check_device_index()`, `device_context_initialized()`. They return empty on a host with no NPU, so `memory_summary()` degrades to a one-line notice instead of raising. The scope line's physical NPU ids come from `_get_device_topology()`, so there is no extra runtime call, and it falls back rather than raising.

### Next Steps

- Build and run against the runtime dependency before undrafting.
- A multi-NPU logical device (`RBLN_NPUS_PER_DEVICE>1`) is still uncovered by the tests.
- `memory_summary()` reports the caching allocator only. `rbln_prof_get_memory()` also counts direct device allocations that `MemoryStats` never sees — worth folding in as a separate total.

---

## Related Issues / Tickets

* Related to #

---

## Type of Change

- [ ] `fix` — Bug fix

---

## Affected Modules

- [ ] Backend
- [ ] Device
- [ ] Ops/Kernels
- [ ] `torch.compile`
- [x] Memory
- [ ] Build/Tools
- [ ] Docs

---

## How to Test

1. `pytest test/internal/test_memory_stats.py -k TestPerChipletMemoryStats`
2. `TestPerChipletMemoryStats` covers the key schema, breakdown-vs-aggregate agreement, the scope line, the no-device notice, and a second logical device. The last one needs at least two logical devices and skips otherwise.
3. `print(torch.rbln.memory_summary())` — one row per (NPU, chiplet), a `total` row, and a scope line naming the pid and the physical NPUs behind the logical device.

---

## Notes

**Runtime dependency.** This needs the paired rebel_compiler change: `rbln_get_memory_stats_per_chiplet()` now returns `[node][chiplet]` instead of a flat vector, and the runtime no longer forwards a torch device id as the allocator's node index. That mismatch is why `memory_stats()` threw for any device other than `rbln:0`, why a multi-NPU logical device reported its first NPU only, and why `empty_cache()` under-released — the stale comment documenting that limitation is removed here. Until the dependency lands, the clang-tidy and build jobs fail against the old flat signature; that is expected, and the PR stays in draft until then.


torch_rbln/memory.py
```diff
@@ -165,6 +173,104 @@ def memory_stats(device: Optional[Union[int, str, torch.device]] = None) -> Dict
...
+    from torch_rbln._internal.rsd_utils import get_physical_device_ids
+
+    physical = get_physical_device_ids(device.index)
+    npus = "[ " + ", ".join(str(pid) for pid in physical) + " ]" if physical else "[unknown]"
+
+    header = f"{'npu':>5}{'chiplet':>9}" + "".join(f"{label:>13}" for _, label in columns)
+    lines = [
+        f"torch_rbln memory summary (device={device}, MiB)",
+        f"scope: pid {os.getpid()}, physical NPU {npus} -- caching allocator only, this process only",
+        header,
```
rebel-chanheo: These IDs are indices into the devices visible to this process, not system-wide physical NPU IDs.
For example, with `RBLN_DEVICES=4,5`, rbln:0 uses physical NPU 4, but this line prints physical NPU [ 0 ], which can mislead users comparing this output with rbln-smi.

Could we label them as visible-device indices, or resolve the actual physical NPU IDs before printing?

rebel-jonghewk: Your right. Applied.

rebel-chanheo (approved):
LGTM

rebel-huijong (approved):
LGTM

cc: @rebel-jongholee

rebel-jonghewk:
@rebel-chanheo Can you merge this? Thanks.

PR #192, opened 2026-08-06
fix(device): defer mapping commit and make availability probes non-throwing

rebel-chanheo:
## Summary of Changes

Availability queries on the RBLN (PrivateUse1) backend become non-throwing and side-effect-free: `torch.rbln.is_available()` / `device_count()`, the accelerator hooks, and everything torch reaches them from. Device mapping splits into *plan* (parse and validate `RBLN_*`, compute the logical→physical table, claim nothing) and *commit* (`rbln_register_device_id`, deferred to first device use). Consumer side of rebel-compiler [#12904](https://github.com/rebellions-sw/rebel_compiler/pull/12904), which moved the runtime's own `RBLN_DEVICES` freeze from the first query to acquisition.

torch calls these from paths that never asked for an NPU — `DataLoader(pin_memory=True)`, `torch.load(map_location=...)`, importing `torch.testing._internal.common_utils`, `torch._utils._get_available_device_type()`. Before this, each of those could raise on a malformed config and each claimed every mapped NPU. Reported as [fsw-inference#495](https://github.com/rebellions-sw/torch-rbln-internal/issues/43) (`import lmcache` alone takes the host's NPUs) and [#475](https://github.com/rebellions-sw/torch-rbln-internal/issues/31).

### Where it lives

| Area | What changed |
|---|---|
| `DeviceMappingManager` | plan / commit split, lazy planning, plan invalidated on `RBLN_*` change, terminal state after a part-way commit |
| `RBLNFunctions` | `get_device_count_nothrow` backs `device_count()`; `device_count_ensure_non_zero` added for the point of use; `to_device_id` is the commit point; `runtime_available` folds in a failed commit |
| `RBLNHooksInterface` | `deviceCount` / `getCurrentDevice` / `setCurrentDevice` / `exchangeDevice` / `maybeExchangeDevice` overridden; probe hooks carry no logging |
| `RBLNGuardImpl` | `deviceCount` delegates to the shared nothrow enumeration |
| `ProcessGroupRBLN` | commits the mapping before starting worker threads — RCCL takes the index directly |
| `Module.cpp`, `device.py` | `device_count()` nothrow, `is_available()` bound to the same predicate as `hasRBLN()` |
| `RBLNLogging.h` | `RBLN_CHECK_QUIET` — throws without logging |

### Plan and commit

Plan runs on any query and claims nothing. Commit runs on first device use — an allocation, `synchronize()`, a collective — and is also where the runtime freezes its own mapping: `rbln_register_device_id` reaches `Context::Create`, which latches it. Both layers therefore freeze in the same call, which is what makes the split safe; a layer that froze at first query while the other froze at commit would mis-map devices silently instead of erroring.

The runtime is the only authority on how many NPUs `RBLN_DEVICES` leaves visible, so this layer never parses it — plan asks `rbln_get_device_count()` and works in indices into whatever pool that returns. Selecting a device with `set_device()` is bookkeeping and does not commit.

Mutating `RBLN_*` concurrently with a query is not supported: before commit a change rebuilds the plan, clearing what a getter reads. Assign the variables from one thread, before the queries that consume them.

### Decisions worth reviewing

| Decision | Why |
|---|---|
| `device_count()` is nothrow; `device_count_ensure_non_zero()` is a separate throwing entry point | `ATen/DeviceAccelerator.h` requires enumeration not to raise, but the config error still has to reach someone. Named after `c10::cuda`'s pair |
| `RBLN_CHECK_QUIET` on the plan path (21 sites); `commit()` keeps loud `RBLN_CHECK` | `RBLN_CHECK` logs `c10::Error::what()` — backtrace included — before throwing, at ERROR level, so a caught availability failure still printed into every co-tenant's console |
| the plan is rebuilt whenever `RBLN_*` changes, until commit | a launcher assigns devices after import. `RBLN_VISIBLE_DEVICES` is in that signature because the runtime treats it as an alias; `RBLN_DUMMY_DEVICE` is not, because the runtime seals it and `is_dummy_device()` caches it |
| a part-way commit is terminal, and reports unavailable | claimed devices cannot be released and `rbln_register_device_id` succeeds for an id it already holds, so a retry against a rebuilt plan would bind this layer to a mapping the runtime does not have |
| `current_device()` clamps to `rbln:0` when the plan no longer holds the selection | the selection is `thread_local` and the plan is process-wide, so a replan can leave one thread past the end. Other threads' selections are unreachable from the one that replans |
| the probe hooks — `isAvailable`, `hasRBLN`, `deviceCount`, `hasPrimaryContext` — do no logging | `RBLN_LOG_DEBUG` can throw; a log after a nothrow predicate makes the C++ hooks able to fail where the Python entry points, bound to the same predicates, cannot |
| dummy mode is not short-circuited in `runtime_available()` | short-circuiting reported available for a dummy device whose mapping had failed to build |

### Contract suite

`test/rbln/test_privateuse1_contract.py` pins one upstream clause per test, citing its source, so a torch upgrade or a newly reached call site fails on the clause rather than on a downstream symptom. Every probe runs in a fresh subprocess and reports whether it raised, opened an NPU context, and whether a later `RBLN_DEVICES` remap is applied or frozen. Positive controls trigger each of those three deliberately, because a clause phrased as "the probe did NOT do X" passes both when the contract holds and when the harness has gone blind.

Three clauses are `strict=True` xfails, all backend surface torch looks for and this one does not expose: `torch.manual_seed()` reaching the RBLN generators, `torch.rbln._utils._get_device_index` for `torch.load`, and `get_device_name` / `get_device_properties` / `get_device_capability`.


rebel-huijong (approved):
So the device seal (rbln_register_device_id) is now deferred until the point where the plan is committed! This was causing a number of problems, I'm really glad to see it fixed so quickly.

PR #194, opened 2026-08-10
feat(profiler): draw the run -> triggered-op correlation as kineto flow arrows

rebel-ywkim:
## Summary of Changes

When device ops are dispatched asynchronously, the device work is not guaranteed to fall within the wall-clock window of the host function that launched it, so a merged `torch.profiler` trace cannot tell which device slices belong to which run. 

The runtime stamps a `launch_id` on the run's host marker and on every device slice it triggered(https://github.com/rebellions-sw/rebel_compiler/pull/12500); this PR turns that into ac2g flow arrows - one source marker per run fanning out to its device slices.

---

## Related Issues / Tickets

* Related to # https://github.com/rebellions-sw/rebel_compiler/pull/12500

---

## Type of Change

- [ ] `fix` — Bug fix

---

## Affected Modules

- [ ] Backend
- [ ] Device
- [ ] Ops/Kernels
- [ ] `torch.compile`
- [ ] Memory
- [ ] Build/Tools
- [ ] Docs

---

1. Run `RBLN_DEVICES=<free device> pytest test/rbln/test_kineto_region_device_alignment.py -q`
   - needs a live NPU and a `rebel-compiler` build that includes [rebel-compiler #12500.
](https://github.com/rebellions-sw/rebel_compiler/pull/12500)
2. Verify output: `4 passed` - device work contained in the launching region under sync dispatch,
   never starting before its launch under async dispatch, the same containment for eager ops, and
   every `rbln_run` marker owning one `ac2g` flow start that reaches device slices of the same
   `launch_id` with nothing dangling.
3. Verify visually: chrome trace in ui.perfetto.dev - a device slice follows the arrow back to its `rbln_run`.
4. Edge cases tested: mutated traces (dropped/re-pointed/stripped arrows) fail the assertions; a launch with no device slice draws no arrow and warns.
---

스크린샷 2026-08-10 오후 12 22 17

---

## Checklist

* [ ] Linting passes — see [Linting](docs/LINTING.md)

---

## Notes

---


rebel-jonghewk (approved):

torch_rbln/csrc/rbln/profiler/kineto/rbln_kineto_emitter.cc
```diff
@@ -25,6 +30,90 @@ ::libkineto::ActivityType kind_to_activity_type(RblnKinetoActivityKind kind) {
...
+}
+
+void add_flow_arrows(const RblnKinetoExport* exp, int64_t clock_offset_ns,
+                     const ::libkineto::TraceSpan& span, ProjectedKinetoTrace* out) {
+  if (exp->host_launches_count == 0) {
+    RBLN_LOG_INFO("rbln flow arrows: no host launches captured; 0 arrows");
+    return;
+  }
+
+  // 1. assign one flow id per host launch.
```
rebel-chanheo, line 65: I'm not very familiar with how Kineto manages flow ID namespaces, so I wanted to check one thing here.

From what I understand, `GenericTraceActivity::Flow` requires IDs to be unique within each flow type, and CUDA also appears to use `kLinkAsyncCpuGpu` with its correlation IDs.
Since the IDs here start from 1 for each export, is there any possibility that they could overlap with IDs from another flow producer?

Perhaps this is already prevented by the CPU-only setup or the profiler lifecycle. Could you clarify whether such coexistence is impossible here, or whether the RBLN flow IDs need their own reserved range/namespace?

References:
- Kineto flow ID uniqueness:
    https://github.com/pytorch/kineto/blob/4f8caaae3f109b63767e75bbb7fae7857e1d3e13/libkineto/include/GenericTraceActivity.h#L182-L190
- CUDA’s use of `kLinkAsyncCpuGpu`:
    https://github.com/pytorch/kineto/blob/4f8caaae3f109b63767e75bbb7fae7857e1d3e13/libkineto/src/CuptiActivity.h#L65-L70
- PyTorch flow-parent handling:
    https://github.com/pytorch/pytorch/blob/v2.11.0/torch/csrc/profiler/collection.cpp

rebel-ywkim: I searched for `kLinkAsyncCpuGpu` producers. kineto has three : CUPTI (`CuptiActivity.h:65-68`), rocprof (`RocprofActivity.h:57-60`) and XPUPTI (`XpuptiActivityHandlers.cpp:203-204`), all using their raw correlation id calculated internally.

PyTorch itself does not produce `kLinkAsyncCpuGpu`. it only consumes ac2g in `collection.cpp` and emits `kLinkFwdBwd` with its own counter. and our wheel compiles all three out (`-DLIBKINETO_NOCUPTI -DLIBKINETO_NOROCTRACER -DLIBKINETO_NOXPUPTI=ON`).

So our flow_id assignment would not conflict with any other producers right now, but I think the way assigning flow_id now itself is fragile in case something new is coming in, especially I was assigning starting from 1.

Since kineto does not provide any precise way to take range/namespace for this case, rather than a reserved namespace we cannot get, as a work-around, can I move our ids into a higher, and practically wide enough field as below? - 

**Layout**
```
[ 11110000  | index(starting from 1)]
       8 bit               24 bit
```

- when flow_id offset exceeds 2^24 - 1, raises  warning and does not create flow_id and Flow.

I will also add assertion on the existing flow-arrow test which is using simple MLP, that every ac2g flow start inside our region maps to one of our rbln_run markers- so anything that torch generates by default emitting ids into our range would be caught.

rebel-chanheo: Thanks for checking more then that and Sound good to me!

rebel-chanheo (approved):
LGTM

rebel-ywkim:
> I searched for `kLinkAsyncCpuGpu` producers. kineto has three : CUPTI (`CuptiActivity.h:65-68`), rocprof (`RocprofActivity.h:57-60`) and XPUPTI (`XpuptiActivityHandlers.cpp:203-204`), all using their raw correlation id calculated internally.
> 
> PyTorch itself does not produce `kLinkAsyncCpuGpu`. it only consumes ac2g in `collection.cpp` and emits `kLinkFwdBwd` with its own counter. and our wheel compiles all three out (`-DLIBKINETO_NOCUPTI -DLIBKINETO_NOROCTRACER -DLIBKINETO_NOXPUPTI=ON`).
> 
> So our flow_id assignment would not conflict with any other producers right now, but I think the way assigning flow_id now itself is fragile in case something new is coming in, especially I was assigning starting from 1.
> 
> Since kineto does not provide any precise way to take range/namespace for this case, rather than a reserved namespace we cannot get, as a work-around, can I move our ids into a higher, and practically wide enough field as below? -
> 
> **Layout**
> 
> ```
> [ 11110000  | index(starting from 1)]
>        8 bit               24 bit
> ```
> 
> * when flow_id offset exceeds 2^24 - 1, raises  warning and does not create flow_id and Flow.
> 
> I will also add assertion on the existing flow-arrow test which is using simple MLP, that every ac2g flow start inside our region maps to one of our rbln_run markers- so anything that torch generates by default emitting ids into our range would be caught.

Applied this. For index, I used 0 based instead of 1 here.

PR #196, opened 2026-08-10
ci: add nightly workflow building against PyTorch nightly

### LINKED ISSUE #195: [CI] build and test against PyTorch nightly on a new daily cron
## Why

We want to enlist `RBLN-SW/torch-rbln` in PyTorch's [Cross-Repository CI Relay (CRCR)](https://github.com/pytorch/crcr-test). CRCR has two prerequisites before an L1 allowlist entry:

1. **Public repo with Actions enabled**: already satisfied. `RBLN-SW/torch-rbln` is public, not archived, with ~2,800 public workflow runs.
2. **CI builds and tests against PyTorch `main`**: not satisfied. The prerequisite reads:

   > On each dispatch **OR a nightly cron**, check out `pytorch/pytorch` at the dispatched commit SHA … Do not pin an old release tag or a vendor fork tip as the upstream source of truth.

We currently pin `torch==2.11.0+cpu` (`pyproject.toml:22`, `pyproject.toml:41`) against the `pytorch-cpu` release index. Nothing in the repo tracks PyTorch `main` or nightly.

I suggest we take the **nightly cron** branch of that clause first. It's the cheap path: PyTorch ships prebuilt nightly CPU wheels, so we never build PyTorch from source, and because the inputs (nightly-of-the-day x `torch-rbln` HEAD) are fixed per day, it's one compile per day.

## Scope

A new workflow that runs daily and on `workflow_dispatch`.

Steps:

1. Resolves the latest `torch` nightly CPU wheel (preferably 00:00 UTC)
2. Builds `torch-rbln` against it
3. Runs the CI smoke suite
4. Reports pass/fail

Must live on the **default branch (`dev`)**.

### 1. Add the nightly index to `pyproject.toml`

`[tool.uv.index]` currently has no nightly entry. Add one (`explicit = true`, so normal resolution never touches it):

```toml
{ name = "pytorch-nightly", url = "https://download.pytorch.org/whl/nightly/cpu", explicit = true },
```

### 2. Repoint the pin in the workflow

`tools/replace_depends.py` already does this. `tools/dev-setup.sh:151` uses the same call shape for `rebel-compiler`:

```bash
python ./tools/replace_depends.py \
  --source torch \
  --target "torch>=2.12.0.dev0" \
  --pyproject-path pyproject.toml \
  --set-index pytorch-nightly
```

### 3. Build and test

- Reuse the build steps from `_build-wheel.yaml`, but without the publish step.
- Echo the resolved `torch` version into the log so staleness is visible.

### 4. CPU only via `RBLN_DUMMY_DEVICE`

`RBLN_DUMMY_DEVICE=1` registers a host-backed stand-in device with no NPU present. Let's use this for now as `test/rbln/test_dummy_device.py` is only a smoke suite. We also do not have to report its green / red status for now.
> You do not need HUD callbacks yet (L2+)

## References

- CRCR: https://github.com/pytorch/crcr-test
- RFC-0050: https://github.com/pytorch/rfcs/pull/90
- Allowlist: https://github.com/pytorch/pytorch/blob/main/.github/allowlist.yml
- Prior-art allowlist PRs: [#186646](https://github.com/pytorch/pytorch/pull/186646) (L1), [#188651](https://github.com/pytorch/pytorch/pull/188651) (L2)

rebel-tskim:
Adds a daily workflow (plus manual dispatch) that resolves the latest PyTorch nightly CPU wheel, repoints the `torch` pin at it, builds the wheel, and smoke-tests it on a host with no NPU via RBLN_DUMMY_DEVICE. This satisfies the outstanding CRCR prerequisite that CI build and test against PyTorch main.

The build steps mirror `_build-wheel.yaml` without publishing to the internal package index. Results are reported to the job summary.

## Summary of Changes

Commit 201ff88 on feat/CRCR-workflow-L1, three files, +268 lines:

**`.github/workflows/nightly-torch.yaml` (new, 245 lines)**
one job, daily at 05:00 UTC plus manual dispatch (python_version, torch_version), TORCH_RBLN_RUNNER_CPU_LARGE in the rbln-build container, owner-gated:

1. Resolve the latest nightly from download.pytorch.org/whl/nightly/cpu, before checkout.
2. Repoint the torch pin with the existing tools/replace_depends.py (no change needed to that script).
3. Build with _build-wheel.yaml's compiler setup and build constraint, without publication.
4. Install the wheel into a clean venv; smoke-test under RBLN_DUMMY_DEVICE=1, then run test_dummy_device.py + test_no_device.py.
5. Report torch / torch-rbln / Python / commit / status to the job summary, with an error annotation on failure.

**`pyproject.toml`**
one line: pytorch-nightly-cpu as an explicit = true index.

**`docs/WORKFLOWS.md`**
table rows plus a section describing the five steps.

## Local test result

Ran the workflow's steps by hand.

| Step | Result |
| --- | --- |
| Resolve latest nightly | ✅ `torch==2.14.0.dev20260809+cpu` |
| Repoint `torch` pin | ✅ both pins + `[tool.uv.sources]` |
| Build against nightly | ❌ 428 compile errors (see below) |
| Build against release pin | ✅ wheel produced |
| Install wheel into clean venv | ✅ |
| Smoke test (`RBLN_DUMMY_DEVICE=1`) | ✅ |

The plumbing works; the nightly build failure is the workflow doing its job. Two causes:

1. `c10/rbln/RBLNPinnedAllocator.cpp:71` uses `std::memcpy` without `<cstring>`, which only compiled because torch 2.11 included it transitively.
2. Nightly headers require C++20, but `CMakeLists.txt:15` sets `CMAKE_CXX_STANDARD 17`. Syntax-only probe of `<torch/library.h>` + `<ATen/ATen.h>`: 16 errors at `-std=c++17`, 0 at `-std=c++20`.

So this workflow will be red until those are fixed.

---

## Related Issues / Tickets

* Resolves #195
* Related to #

---

## Type of Change

- [ ] `fix` — Bug fix

---

## Affected Modules

- [ ] Backend
- [ ] Device
- [ ] Ops/Kernels
- [ ] `torch.compile`
- [ ] Memory
- [x] Build/Tools
- [x] Docs

## Checklist

* [x] This PR is linked to a corresponding issue — see [Contributing to torch-rbln](docs/CONTRIBUTING.md)

### DIFF
--- a/.github/workflows/nightly-torch.yaml
+++ b/.github/workflows/nightly-torch.yaml
@@ -62,17 +62,21 @@ jobs:
         run: |
           set -euo pipefail
 
+          requirement=torch
           if [[ -n "${TORCH_VERSION}" ]]; then
-            target_version="${TORCH_VERSION}"
-          else
-            target_version="$(printf 'torch\n' \
-              | uv pip compile - --no-config --no-deps --no-header --no-annotate \
-                  --prerelease allow --default-index "${PYTORCH_NIGHTLY_INDEX_URL}" \
-              | sed -n 's/^torch==//p')"
-            if [[ -z "${target_version}" ]]; then
-              echo "::error::Failed to resolve torch from ${PYTORCH_NIGHTLY_INDEX_URL}"
-              exit 1
-            fi
+            requirement="torch==${TORCH_VERSION}"
+          fi
+
+          # Resolving even an explicit version normalizes it to what the index actually
+          # serves (`2.14.0.dev20260810` -> `2.14.0.dev20260810+cpu`), which the smoke
+          # test compares against `torch.__version__`.
+          target_version="$(printf '%s\n' "${requirement}" \
+            | uv pip compile - --no-config --no-deps --no-header --no-annotate \
+                --prerelease allow --default-index "${PYTORCH_NIGHTLY_INDEX_URL}" \
+            | sed -n 's/^torch==//p')"
+          if [[ -z "${target_version}" ]]; then
+            echo "::error::Failed to resolve ${requirement} from ${PYTORCH_NIGHTLY_INDEX_URL}"
+            exit 1
           fi
 
           echo "target_version=${target_version}" >> "${GITHUB_OUTPUT}"
--- a/.github/workflows/nightly-torch.yaml
+++ b/.github/workflows/nightly-torch.yaml
@@ -110,7 +110,6 @@ jobs:
           UV_INDEX_RBLN_USERNAME: ${{ secrets.INTERNAL_INDEX_USERNAME }}
           UV_INDEX_RBLN_PASSWORD: ${{ secrets.INTERNAL_INDEX_PASSWORD }}
           UV_BUILD_CONSTRAINT: constraints-build-dev.txt
-          UV_PRERELEASE: allow
         run: |
           set -euo pipefail
 
@@ -149,7 +148,6 @@ jobs:
           UV_INDEX_STRATEGY: unsafe-best-match
           UV_INDEX_RBLN_USERNAME: ${{ secrets.INTERNAL_INDEX_USERNAME }}
           UV_INDEX_RBLN_PASSWORD: ${{ secrets.INTERNAL_INDEX_PASSWORD }}
-          UV_PRERELEASE: allow
         run: |
           set -euo pipefail
 
--- a/.github/workflows/nightly-torch.yaml
+++ b/.github/workflows/nightly-torch.yaml
@@ -1,0 +1,245 @@
+name: Nightly PyTorch
+
+on:
+  schedule:
+    # Daily 14:00 KST (05:00 UTC)
+    - cron: 0 5 * * *
+  workflow_dispatch:
+    inputs:
+      python_version:
+        description: Python version (3.10-3.13)
+        type: string
+        required: false
+        default: '3.12'
+      torch_version:
+        description: PyTorch nightly version to pin (leave empty to resolve the latest nightly)
+        type: string
+        required: false
+        default: ''
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: false
+
+defaults:
+  run:
+    shell: bash
+
+env:
+  PYTORCH_NIGHTLY_INDEX_NAME: pytorch-nightly-cpu
+  PYTORCH_NIGHTLY_INDEX_URL: https://download.pytorch.org/whl/nightly/cpu
+  TEST_VENV: nightly-test/.venv
+
+jobs:
+  nightly_torch:
+    name: Build and test against PyTorch nightly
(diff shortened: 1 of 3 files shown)


.github/workflows/nightly-torch.yaml
```diff
@@ -0,0 +1,245 @@
...
+          grep -nE '"torch==|^torch = ' pyproject.toml
+
+      - name: Build wheel
+        env:
+          TORCH_RBLN_BUILD_TYPE: Release
+          UV_INDEX: rbln=${{ secrets.INTERNAL_INDEX_URL }}
+          UV_INDEX_RBLN_USERNAME: ${{ secrets.INTERNAL_INDEX_USERNAME }}
+          UV_INDEX_RBLN_PASSWORD: ${{ secrets.INTERNAL_INDEX_PASSWORD }}
+          UV_BUILD_CONSTRAINT: constraints-build-dev.txt
+          UV_PRERELEASE: allow
```
rebel-chanheo: Do we need `UV_PRERELEASE=allow` for the entire build and install steps?

The torch dependency is already rewritten to an exact nightly version, and the repository's uv configuration allows explicitly pinned prereleases.
Applying `allow` globally may also select prerelease versions of unrelated dependencies.

Could we remove this override here and in the install step, and verify the workflow still resolves? If a transitive prerelease dependency requires it, please pin that dependency explicitly or document why the global override is necessary.

.github/workflows/nightly-torch.yaml
```diff
@@ -0,0 +1,245 @@
...
+      # cannot influence which wheel is picked.
+      - name: Resolve PyTorch nightly version
+        id: resolve_torch_version
+        env:
+          TORCH_VERSION: ${{ inputs.torch_version }}
+        run: |
+          set -euo pipefail
+
+          if [[ -n "${TORCH_VERSION}" ]]; then
+            target_version="${TORCH_VERSION}"
```
rebel-chanheo: Non-blocking: Would it be worth resolving or normalizing a manual `torch_version` before using it as `target_version`?

Concern is that an input such as `2.14.0.dev20260810` may resolve to `2.14.0.dev20260810+cpu`. The build and installation would succeed, but the later exact comparison against `torch.__version__` could make the manual run fail.

If callers are expected to always include the full local suffix such as `+cpu`, documenting or validating that expectation would also address this.

.github/workflows/nightly-torch.yaml
rebel-tskim: Good catch! Removed UV_PRELEASE=allow flag from both steps. Local build showed no difference.

.github/workflows/nightly-torch.yaml
rebel-tskim: Added normalization logic to regard 2.14.0.dev20260810 as 2.14.0.dev20260810+cpu.

rebel-chanheo (approved):
LGTM

rebel-tskim:
I will wait for tomorrow's actual run at 2PM and proceed to upstream side work. Thanks!

PR #197, opened 2026-08-10
feat(copy): batch cpu<->rbln pairs in _foreach_copy_ via h2v/v2h multi-copy

rebel-chanheo:
## Summary of Changes

`aten::_foreach_copy_` on rbln reaches the runtime once per pair. `lmcache-rbln`'s KV offload — the only in-tree caller — builds one pair per (block, layer, half, head), so storing a 1024-token chunk is 1984 separate submits.

rebel-compiler#12439 ships `rbln_memcpy_h2v_multi` / `v2h_multi`: one call carrying many slab descriptors. This PR makes them reachable from the torch layer and routes `_foreach_copy_`'s **contiguous** host-crossing pairs through them, so a list reaches the runtime once per bulk call instead of once per pair.

Only `aten::_foreach_copy_` changes behaviour, and only for contiguous pairs crossing the host boundary. `rbln→rbln` across devices stays ineligible as before; host-direction lists spanning devices stay batched, one bulk submit per device.

**Two of the changes here are not the feature.** Both sit in `_foreach_copy_`'s reorder guard, and the feature cannot land without them.

The guard tracked device tensors only, so a device→host list had no disjointness check at all — and the runtime requires disjoint destinations without validating them, which makes this function the only defence. Not a live bug on `dev`, which does not batch host pairs, but it has to exist before the batch does.

The same guard compared every pair against every other pair, and it runs on every call whether or not anything gets batched. So `dev` pays it today: 6.5 ms at 2048 pairs, 25 ms at 4096, before a byte moves. Same predicate now, over sorted ranges with a binary search.

**One bulk call cannot carry everything, and overshooting is not recoverable.** Past 16 entries / 8 MiB the submit never returns: the device raises `SYS_KERNEL_TIMEOUT` and cancels the context, after which the per-entry retry *and* the CPU fallback fail too. So the cap has to prevent the case rather than react to it. Measured on lmcache's store, 24 entries / 12 MiB pass and 32 / 16 MiB start timing out, so the cap sits a factor of two under.

Two consequences follow. The guarantee is one submit per 16 entries / 8 MiB **per device**, not one submit per call. And a single descriptor over 8 MiB cannot be split, so it takes the single-copy path, which has no such limit. The runtime API documents no limit, so this is a defect — the cap should be deleted once it is fixed. v2v is unaffected.

### Results

Store path (`gather_full_blocks_headmajor`, device→host), L=62 H=8 D=128 block_size=512 bf16, median of 3 runs:

| chunk | pairs | dev | this PR | submits |
|---|---|---|---|---|
| 512 | 124 | 9.70 ms | 9.42 ms (−3%) | `v2h` 124 → `v2h_multi` 16 |
| 1024 | 1984 | 37.74 ms | 35.88 ms (−5%) | `v2h` 1984 → `v2h_multi` 124 |
| 2048 | 3968 | 75.53 ms | 71.71 ms (−5%) | `v2h` 3968 → `v2h_multi` 248 |

The restore path is **unchanged** by this PR: its pairs are strided (the scatter's source is a `[H, block_size, D]` view of a wider chunk), and this PR batches contiguous pairs only. Describing those pairs instead of staging them is rebellions-sw/fsw-inference#199, which builds on the classes added here.

---

## Related Issues / Tickets

* Related to rebellions-sw/torch-rbln-internal#30 — payoff (d)
* Related to rebel-compiler#12439 — the `*_multi` entrypoints this routes through

---

## Type of Change

- [x] `fix` — Bug fix

---

## Affected Modules

- [x] Backend
- [ ] Device
- [ ] Ops/Kernels
- [ ] `torch.compile`
- [x] Memory
- [ ] Build/Tools
- [ ] Docs

---

## How to Test

1. `python -m pytest test/rbln/test_foreach_copy_host.py test/rbln/test_foreach_copy_v2v.py -q`
   - Expect **58 passed**. Batching is asserted through the `torch.rbln.explain()` rt-timing
     counters, not the values — a value-only test passes on an unbatched implementation:

     ```
     N contiguous pairs, one direction, under the cap  →  *_multi 1 call, per-entry 0
     40 × 1 MiB, one direction                        →  5 calls (the byte cap)
     40 × 64 KiB, one direction                       →  3 calls (the entry cap)
     one pair over 8 MiB                              →  *_multi 0, single copy 1
     non-contiguous pair / pinned + non_blocking      →  *_multi 0, values still correct
     ineligible pair between two v2v pairs            →  v2v_multi 1, not 2
     mixed directions                                 →  3 submits total, not one per pair
     happy path                                       →  0 bounces
     ```
2. `./build/bin/test_cpp_RBLNHostBatchTest && ./build/bin/test_cpp_RBLNV2VBatchTest`
   - Expect **19 + 18 passed**, including cross-device splitting and buffer lifetime.
3. `python -m pytest test/rbln/test_cat_index_select_v2v.py test/rbln/test_copy_v2v.py test/rbln/test_index_copy_v2v.py test/rbln/test_tensor_copy.py test/rbln/test_profiler.py -q`
   - Expect **1190 passed, 6 skipped** — regression sweep over the v2v consumers and the
     profiler.

---

## Notes

Three things worth weighing separately:

* **What this PR buys on its own** — the store path above, plus removal of an O(n²) guard
  that `dev` pays on every `_foreach_copy_` call regardless of batching.
* **What it is the substrate for** — `H2VBatch` / `V2HBatch` / `enqueue_strided` are what
  rebellions-sw/fsw-inference#199 uses to take the restore path 3.7x. That cannot land without these.
* **What it costs** — a pure-v2v `_foreach_copy_` pays `+0.10 / +0.14 / +0.20 µs` at N=1/2/4
  (6–7%) for the per-pair guards and the two unused host batches. No single addition accounts
  for it. No in-tree caller produces that shape — lmcache's lists are host-direction.

Deferred on purpose:

* Route the `copy_` staging paths through the same primitives and batch strided host pairs
  there — rebellions-sw/fsw-inference#199, where the op actually produces them and the per-descriptor measurement can
  justify a threshold.
* Lift the 16-entry / 8 MiB cap once the runtime timeout is fixed.
* Put pinned + `non_blocking=True` pairs back on the batch once an async multi variant
  exists — batching them today would trade a real overlap for a submit count.

---

## Checklist


rebel-huijong (approved):
As a side effect of this [commit](https://github.com/RBLN-SW/torch-rbln/pull/197/changes/ee7d8ddf39ecb2527d667ea27f33df50f0367118), the existing last-write-wins behavior is now blocked for v2v as well. This is a point where we diverge from GPU, so I think it's something we should all be aware of. Since _foreach_copy_ is a private API on the PyTorch side anyway and isn't really documented, I don't think we need to document it on our end either.

rebel-chanheo:
> As a side effect of this [commit](https://github.com/RBLN-SW/torch-rbln/pull/197/changes/ee7d8ddf39ecb2527d667ea27f33df50f0367118), the existing last-write-wins behavior is now blocked for v2v as well. This is a point where we diverge from GPU, so I think it's something we should all be aware of. Since foreach_copy is a private API on the PyTorch side anyway and isn't really documented, I don't think we need to document it on our end either.

Right that it's now blocked for v2v too. On the GPU divergence though — CPU _foreach_copy_ raises on the same input, so the old silent behaviour looks like it was ours (our batch wasn't running copy_'s overlap check) rather than PyTorch's. I haven't run CUDA, so if it really does last-write-wins there, please correct me.

rebel-huijong:
That's how I remember it, but I'll do a quick check on GPU next time I'm working there and share what I find.

PR #198, opened 2026-08-10
refactor(core): locate librbln.so from the rebel-compiler record instead of TVM

rebel-chanheo:
## Summary of Changes

`librbln.so` was located by a vendored copy of `tvm/_ffi/libinfo.py`, and CMake hard-coded `<site-packages>/tvm`. rebel-compiler is removing TVM. The vendored copy had also drifted from its original: it walked every `PATH` and `LD_LIBRARY_PATH` entry recursively, cost 65 ms per import, and let any `librbln.so` under e.g. `/usr/bin` outrank the installed wheel.

**Resolution.** The library is the one the `rebel-compiler` distribution recorded installing, anchored on the `rebel` package this interpreter imports. Recorded paths are relative to the install root, so a single lookup answers the three questions the build and the runtime used to answer separately — where the library is, where it sits relative to the package (what the install RPATH is written against), and where the headers are. It follows a relocation with no change here. ~18 ms.

**Override.** `LD_LIBRARY_PATH`, and only that. The dynamic loader and rebel-compiler's loader both honour it, so all three sides land on the same file. `TVM_LIBRARY_PATH` is neither read nor written: it is TVM's variable, and an override only one side honours cannot make the sides agree.

**Ordering.** No layout is guessed ahead of the record — a guess can only shadow it, for exactly as long as a stale copy sits at the old path, which is the state a relocation leaves behind. The single fallback, `<source>/build`, runs only when the record does not name the library; that is the editable install it exists for.

**Loading.** The library is mapped before the native extensions. They declare it NEEDED by SONAME, so they reuse that mapping rather than searching a RUNPATH baked in at build time — which is how a library that moved after those binaries were built still resolves. An existing mapping is adopted, so a program that imported rebel-compiler first keeps the copy it has.

**Build.** `tools/find_rebel_runtime.py` runs the same resolver; the install RPATH and the include directory come from recorded entries. Today that still yields `tvm/`, so binaries are unchanged. FindRebel now also uses the interpreter the build was given rather than one discovered on `PATH`, which may belong to a different environment. The wheel job verifies twice, because one step was answering two questions at once. Before publishing it installs `dist/*.whl` and loads the extension with `TORCH_DEVICE_BACKEND_AUTOLOAD=0` — autoload maps `librbln.so` through `torch_rbln` first and leaves the RUNPATH untested — so a bad artifact never reaches the index; publishing used to come first, and the publish step skips a filename already there, so a re-run could not replace it. After publishing it installs `torch-rbln==<version>` from the index, which the local check cannot cover: that the upload landed and the version is resolvable by name.

Out of scope: released wheels bake `$ORIGIN/../../tvm` into RUNPATH, so when rebel-compiler moves the library it must leave a **symlink** at the old path (a copy has a different inode and would load a second runtime) and keep the `librbln.so` SONAME.

---

## Type of Change

---

## Affected Modules

- [x] Backend
- [x] Build/Tools

---

## How to Test

1. `python -m pytest test/rbln/test_rbln_runtime_lib.py` — 11 tests: resolution order, anchoring on the imported `rebel` package, a relocated library, a leftover build tree losing to the record, the editable fallback, and what the build takes from the same record.
2. `python -m torch_rbln.diagnose` — reports the mapped library, its version, and every candidate considered with the rule that produced it.
3. RUNPATH is unchanged: `readelf -d torch_rbln/lib/libtorch_rbln.so | grep RUNPATH` → `$ORIGIN:$ORIGIN/../../torch/lib:$ORIGIN/../../tvm`.
4. The RPATH is genuinely exercised (a plain import never consults it): `TORCH_DEVICE_BACKEND_AUTOLOAD=0 python -c "import ctypes, torch; ctypes.CDLL('<site-packages>/torch_rbln/lib/libtorch_rbln.so')"`.
5. `LD_LIBRARY_PATH=<dir with librbln.so> python -c "import torch_rbln"` loads that library.
6. `docs/WORKFLOWS.md` describes the order the wheel job now runs in.

---

## Checklist

---

## Notes

`torch_rbln/_internal/tvm_libinfo.py`, `test/rbln/test_find_and_load_tvm_library.py` and `torch_rbln.find_and_load_tvm_library` are removed. No environment variable is introduced.

torch-rbln no longer reads, writes, imports or hard-codes anything TVM. The `tvm` that still appears in RUNPATH is the directory rebel-compiler records installing into, i.e. data from the record rather than a decision here; it changes when they move the library.

`torch_rbln/_internal/rbln_runtime_lib.py` is loaded by path from `tools/find_rebel_runtime.py` during the build, before torch_rbln exists, so it stays stdlib-only and answers only what the build and the loader ask. Diagnostics-only code lives in `env_diagnostic.py`, which queries the resolver rather than restating its decisions.

diagnose reports one runtime library, not the pair rebel-compiler ships. `librbln_runtime.so` is a runtime-only build its loader can take *instead* of `librbln.so`, and the two cannot both be mapped — loading the second aborts the process in its constructors. There is no state in which reporting both would say anything the abort has not.

**Deliberately not in this PR.** An earlier revision also added single-runtime and version enforcement — duplicate-mapping detection over `/proc/self/maps`, a wrapper forcing every rebel import through a check, a `BaseException`-derived error, refusal of a replaced mapping, and a build-stamped SONAME with the packaging changes to ship it. That is a separate policy with its own failure modes (two were found in review: a conflict bypassing the entry point's status rollback, and an unreadable `/proc/self/maps` reported as a missing library), and the one divergence that could actually be reproduced — `TVM_LIBRARY_PATH` pointing at a second copy — aborts inside `ld.so` before any Python-level check runs. It belongs in its own PR, with a case that shows it catching something.


torch_rbln/_internal/rbln_runtime_lib.py
```diff
@@ -0,0 +1,316 @@
...
+from importlib.metadata import distribution, PackageNotFoundError, version as _dist_version
+
+
+RUNTIME_LIB_NAME = "librbln.so"
+_COMPILER_DIST_NAME = "rebel-compiler"
+
+# Known install layouts, relative to the parent of the rebel package. Checked before the
+# distribution file list because a stat is far cheaper than parsing it. "tvm" is TVM-only; the
+# file list keeps covering it either way.
+_LAYOUT_HINTS = ("tvm", os.path.join("rebel", "lib"))
```
rebel-jongho: is this line still needed in this PR??

rebel-chanheo: No — removed. Both were guesses that could only shadow the distribution record.

.github/workflows/_build-wheel.yaml
```diff
@@ -211,3 +183,63 @@ jobs:
               )
           print(f"::notice::Verified torch-rbln=={torch_rbln.__version__} ({build_type})")
           PY
+
```
rebel-chanheo, line 186: @rebel-clee PTAL at the workflow changes, especially the pre-publish wheel and RPATH verification.

rebel-huijong (approved):
LGTM

.github/workflows/_build-wheel.yaml
rebel-clee: Publishing only what passed makes sense to me.

One thing that went with it, though: we don't install from the index anywhere now. Could we bring that back as a step after publishing?

`docs/WORKFLOWS.md` still says the workflow "publishes it, and verifies it installs in a clean environment" — might be worth updating that too.

rebel-chanheo: Both fixed in 0332343 — index install is back as a post-publish step (it only imports; the artifact is verified before publishing), and WORKFLOWS.md nowdescribes the actual order.

.github/workflows/_build-wheel.yaml
```diff
@@ -139,6 +139,83 @@ jobs:
...
+          already = mapped_runtime_libs()
+          if already:
+              raise SystemExit(f"::error::librbln.so was already mapped ({already}); RPATH untested")
+
+          library = str(next(f for f in distribution("torch-rbln").files if f.name == "libtorch_rbln.so").locate())
+          try:
+              ctypes.CDLL(library)
+          except OSError as e:
+              raise SystemExit(f"::error::install RPATH does not resolve librbln.so: {e}") from e
+          print(f"::notice::Verified the install RPATH of {library} -> {mapped_runtime_libs()}")
```
rebel-clee: The check still passes with a wrong RUNPATH — any `librbln.so` on the search path satisfies the dependency. Worth asserting the resolved path matches the record? `distribution("rebel-compiler")` is the safer way to read it, since importing `rbln_runtime_lib` would run `torch_backends_entry_point()`.

rebel-chanheo: Fixed in 7e8579a — asserts the mapped path is the recorded one, read via `distribution("rebel-compiler")` as you suggested.

rebel-clee (approved):
LGTM

torch_rbln/_internal/rbln_runtime_lib.py
```diff
@@ -0,0 +1,216 @@
+"""Locating and loading the RBLN runtime shared library (``librbln.so``).
+
+This was a vendored copy of ``tvm/_ffi/libinfo.py``, so torch-rbln followed TVM's directory list
+to land on the file rebel-compiler loads. That coupling goes away as rebel-compiler drops TVM,
+and the copy had drifted: it walked every ``PATH`` and ``LD_LIBRARY_PATH`` entry recursively.
```
rebel-jongho: nit: It doesn't seem necessary to include a comment about the history of this file.

rebel-jongho (approved):

torch_rbln/_internal/rbln_runtime_lib.py
rebel-chanheo: Dropped, and did the same pass over the rest.

PR #203, opened 2026-08-11
test: re-derive the logit tolerance for the current precision policy

rebel-chanheo:
## Summary of Changes

The target's compute precision policy changed, so the previously calibrated logit
tolerance no longer reflects the accuracy the platform provides. Re-derived it.

Also adds `test/rbln/test_fp16_numerics.py` — a few-second numerical accuracy check
to run before a model-level accuracy test, so a shift is visible without a model
download and a 90s run. Its bounds are empirical for the current target and are
meant to be re-derived alongside the model-level ones.

Constraints pin moved to `0.11.2.dev256+gc5eaca9d.prod`.

---

## Type of Change

- [x] `test` — Test-only change

---

## Affected Modules

- [x] Tests

---

## How to Test

```bash
pytest test/rbln/test_fp16_numerics.py -q
# 5 passed in ~3s

pytest "test/models/test_transformers.py::TestCausalLMPRIVATEUSE1::test_qwen2_sdpa_b2_s1024_rbln_float16" -q
# 1 passed in ~90s
```

Both verified locally against the pinned version on RBLN-CR03.

---

## Checklist

* [ ] All CI tests pass — pending

### DIFF
--- a/constraints-build-dev.txt
+++ b/constraints-build-dev.txt
@@ -1,2 +1,2 @@
 # Build/dev dependency constraints (single source of truth).
-rebel-compiler==0.11.2.dev123+g5a1ee442.prod
+rebel-compiler==0.11.2.dev256+gc5eaca9d.prod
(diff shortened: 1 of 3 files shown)


rebel-huijong (approved):
LGTM

PR #206, opened 2026-08-21
test: stop tying device-free tests to device availability

rebel-chanheo:
## Summary of Changes

`instantiate_device_type_tests` deletes the template class and rebuilds it per device type, and the RBLN base joins that list only when a device is available. On a host without an NPU the template is gone and nothing replaces it, so the file collects nothing — no failure, no skip.

Three files paid that for nothing. The ABI window checks, the `librbln.so` resolution rules, and the arch predicates use no device and had no axis to expand, so the call only renamed their classes while making them disappear wherever there is no hardware — which is where the ABI and resolution paths most want testing. They are plain `TestCase` now.

`docs/TEST_GUIDE.md` §5 said the call belongs at the bottom of every test file. It now states what the mechanism does and when not to reach for it.

---

## Related Issues / Tickets

* Related to #

---

## Type of Change

- [x] `docs` — Documentation
- [x] `other` — Test-only change (no production code)

---

## Affected Modules

- [x] Docs

---

## How to Test

```bash
uv run --no-sync pytest test/rbln/test_abi_check.py test/rbln/test_rbln_runtime_lib.py test/internal/test_device_arch.py
```

The same tests collect and pass as before the change.

---

## Checklist

* [x] Relevant documentation has been updated

---

## Notes

No production code changes. The two remaining mixed files (`test_device_mapping.py`, `test_memory_stats.py`) touch a device in part and are left alone; splitting them is a separate change.

### DIFF
--- a/.claude/skills/adding-an-op/SKILL.md
+++ b/.claude/skills/adding-an-op/SKILL.md
@@ -1,0 +1,78 @@
+---
+name: adding-an-op
+description: Use when adding, changing, or removing an aten operator on the RBLN backend — the three registration paths, which one a given op belongs on, and what else has to change once you pick.
+---
+
+# Adding an operator
+
+An op reaches the RBLN backend by one of three routes. Picking the wrong one produces code that works locally and then disappears, or an op whose tests never run.
+
+## 1. Pick the path
+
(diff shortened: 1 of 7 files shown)


rebel-chanheo:
Opened from the wrong branch — this carries the agent-conventions changes, not the test-guide ones. Superseded by #207.

PR #209, opened 2026-08-21
docs(agent): write down the conventions for agent-written changes

rebel-chanheo:
## Summary of Changes

Nothing in the repository said which conventions a change has to follow, so every coding agent guessed and reviewers wrote the same comments by hand. The shapes that came back were consistent: verbose comments, an abstraction for one call site, a workaround where the cause was reachable, and tests that pass whether the code is right or wrong.

`AGENTS.md` holds the rules that apply to any change; `CLAUDE.md` is a one-line pointer to it. The steps that only matter at one moment live in `.claude/skills/` so they stay out of the always-loaded file: `adding-an-op`, `writing-tests`, `debugging`, `finishing-a-change`.

Most of the content is repository facts an agent cannot infer and reviewers have had to repeat:

- A change under `aten/`, `c10/`, `torch_rbln/csrc/`, or `tools/codegen/` is not in the interpreter until a rebuild — every test result before that describes the previous build.
- `torch_rbln/_internal/register_ops.py` is generated and gitignored, so fixing an op there works until someone builds.
- `rebel-compiler` is a layer this repository does not own. A guard that hides a compiler defect hides it from the team that can fix it.
- Matching numbers do not show that an op ran on the device: an unaligned or fp32 input takes the host path and then agrees with CPU exactly.
- An op registered only in C++ is invisible to op-test discovery unless it is listed in `test/filters.py`.

Every rule is written so a reviewer can point at it and say yes or no, and the sections that pull against each other — fix the cause, keep the diff reviewable — say so rather than pretending they do not.

---

## Related Issues / Tickets

* Related to #

---

## Type of Change

- [x] `docs` — Documentation

---

## Affected Modules

- [x] Docs

---

## How to Test

1. `uv run --no-sync lintrunner -m origin/dev`
2. Spot-check the facts against their sources: `test/conftest.py` for the fixtures and markers, `tools/codegen/` for what the build generates, `torch_rbln/csrc/rbln/DispatchShim.h` for the shim contract, `docs/CONFIGURATION.md` for the ABI handshake.

---

## Checklist

---

## Notes

Docs only; no production code and no test changes.

`.gitignore` also excludes the per-machine agent settings that sit beside the skills, so only the shared files are tracked.

### DIFF
--- a/.claude/skills/finishing-a-change/SKILL.md
+++ b/.claude/skills/finishing-a-change/SKILL.md
@@ -74,6 +74,7 @@ General:
 - **`else` branches for cases that cannot happen** — should be `raise`, `assert`, or `RBLN_CHECK`
 - **Dead leftovers** — renamed `_var`, unused re-exports, `# removed` comments
 - **Anything not in English**
+- **Absolute measurements** — throughput, latency, memory, or accuracy from an internal run. Take them out; this repository is public. A relative change is fine.
 
 torch-rbln specific:
 
@@ -107,6 +108,7 @@ An empty list is a claim. Only write it if it is true.
 - You are on a feature branch off `dev`, not on `dev` or `main` itself
 - PR title in `type(scope): summary` — CI enforces it. Individual commits only need to be readable (`docs/CONTRIBUTING.md`)
 - Explain intent, not the diff. Symptom, reproducer, where to look, what you verified — the debugging path that got you there is not part of it
+- A perf change may give a relative change, never an absolute measurement
 - Say which layer you changed and which execution paths you exercised — eager, `torch.compile(backend="rbln")`, or both
 - Fill in **Affected Modules** and **How to Test** in the template, and link the issue
 - English
--- a/.claude/skills/finishing-a-change/SKILL.md
+++ b/.claude/skills/finishing-a-change/SKILL.md
@@ -1,0 +1,112 @@
+---
+name: finishing-a-change
+description: Use before claiming a change is done, fixed, ready, or passing, and before writing a commit or PR — rebuilds, runs lint and the tests, states the regression and performance impact, reviews the diff against the repository rules, and reports what was not done.
+---
+
+# Finishing a change
+
+By the time you get here the session is long and the instructions you read at the start have faded. Work through this list rather than from memory.
+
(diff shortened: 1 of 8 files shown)


.claude/skills/finishing-a-change/SKILL.md
```diff
@@ -0,0 +1,112 @@
...
+- **An op registered only in C++** is listed in `_ops_with_rbln_native_kernel` in `test/filters.py`.
+- **A skill whose content you changed still matches its `description`** — the frontmatter is the trigger, so anything the description does not mention may never be loaded.
+- **New or changed tests follow `docs/TEST_GUIDE.md`** — placement, template shape, marker, and the existing fixtures and helpers rather than new ones. Re-read the relevant section against your diff; this is checked in review.
+- **New test files** carry `# Owner(s): ["module: PrivateUse1"]` and the `run_tests()` main block, and are marked `test_set_ci` unless you can say why not.
+- **No widened tolerance** without a derivation, and no test made green by re-running it.
+- **Nothing on the import path uses a device** — the first allocation, `synchronize()`, or collective commits the mapping and freezes it for the process.
+- **PR body claims match the diff.** A conditional check is not described as unconditional; a derived set is not described as derived if it is hardcoded.
+- **`native_functions.yaml` and `test/ops/test_ops.py`** stay close to upstream.
+
+## 6. Report what you did not do
```
rebel-jonghewk, line 95: Can you also consider adding "What not to publish". The repository is public, so throughput, latency, memory and accuracy form internal runs stay out of code, commits and PR descriptions. (https://github.com/RBLN-SW/vllm-rbln/pull/959)

rebel-chanheo: Added — same wording as vllm-rbln#959, plus one line for `docs/CONTRIBUTING.md` (methodology in the PR, numbers not), and fixed the one existing case in `docs/EXPLAIN.md` by dropping the attribution while keeping the example numbers.

rebel-chanheo:
It has been pending for too long, so I am planning to merge it unless you have any specific suggestions. Please feel free to leave a comment!

rebel-huijong (approved):

PR #213, opened 2026-08-26
feat(memory): expose device-memory binding and huge host buffers

rebel-chanheo:
## Summary of Changes

Two APIs for a consumer that DMAs between RBLN device memory and host memory without going through a torch op — so each side has to be ready before the transfer, not materialized during it. The caller is LMCache-RBLN's direct-storage (NVMe) offload path, which reaches past torch-rbln into `rebel._C` for both today (rebellions-sw/fsw-inference#526).

Layer 1 (`torch_rbln/`) and layer 2 (`c10/`, `torch_rbln/csrc/`). Eager dispatch only.

### `torch.rbln.bind_device_memory(tensor)`

An allocation reserves a virtual address and materializes the physical memory behind it on first use through a torch op. A consumer reading those physical buffers out of band never runs one, so it finds nothing there.

The collective path already solves this per operation — `ensureRcclRawMemory` calls `rbln_set_raw_memory_alloc` before every collective, from a dozen sites. This lifts that call into `c10::rbln::bind_device_memory`, routes `ProcessGroupRBLN`'s two binding sites through it, and exposes it:

```python
staging = torch.empty(nbytes, dtype=torch.uint8, device="rbln:0")
torch.rbln.bind_device_memory(staging)
```

`TORCH_RBLN_EAGER_MALLOC`, `is_eager_malloc()`, and the lazy/eager split in `c10::rbln::malloc` are untouched — this binds one region, it does not add an allocation mode.

The tensor has to cover its whole storage: an interior address (a slice, a narrowing view) would configure the wrong region, while a whole-storage `base.view(...)` carries the allocation's own address and size and is accepted. `ensureRcclRawMemory` keeps resolving `key_vaddr` through `rbln_get_memory_info` because it is handed interior views.

### `torch.rbln.huge_host_empty(nbytes)`

The host side of the same transfers, as a tensor over `rebel.host_memory.HugeHostBuffer` (rebel_compiler#12443) — 2 MiB aligned, prefaulted, released with the tensor's last reference. It asks for transparent huge pages through `madvise`, which is best effort: with THP disabled the alignment still holds and nothing is huge-page backed. It replaces the `allocate_huge_host_memory` / `free_huge_host_memory` pair #12443 deprecated, leaving that pair with no caller.

An ordinary CPU tensor is 64 B aligned, so `MultiCopyCmdBuffer` stages it through a bounce buffer (`multi_copy_cmd_buffer.cc`, against `kHostPageSize`); and even an aligned buffer costs a page fault per 4 KiB the first time the runtime resolves its host addresses, inside the transfer rather than in setup.

**`RBLNPinnedAllocator` is deliberately untouched.** It already returns page-aligned memory, which is all the bounce check looks at, so raising its alignment would change no transfer. Whether every `pin_memory=True` allocation should get huge pages is a separate question needing its own measurement.

---

## Related Issues / Tickets

* Related to rebellions-sw/fsw-inference#526 — the LMCache cleanup these let land
* Related to rebellions-sw/fsw-inference#428 — the linear-stack invariant
* Consumes rebellions-sw/rebel_compiler#12443, whose follow-up list names this

## Type of Change

## Affected Modules

- [x] Memory

---

## How to Test

```bash
uv run --no-sync pytest test/rbln/test_tensor_memory.py -k "BindDeviceMemory or HugeHostEmpty" -v
```

**All 12 pass on hardware** (built against the pinned `rebel-compiler==0.11.3.dev0`):

* `TestBindDeviceMemory` — the bind puts physical memory behind a lazy allocation (below); the region stays usable afterwards; re-binding succeeds and does not allocate again (the collective path re-binds before every operation, so it must); a CPU tensor and an interior view are rejected, a whole-storage view is not; the call raises rather than faulting once the runtime is shutting down.
* `TestHugeHostEmpty` — returns a `uint8` CPU tensor of the requested size, 2 MiB aligned; the buffer outlives its `HugeHostBuffer` handle; zero, negative, non-integer and out-of-range sizes raise. Host-only, no NPU needed.

Two of them discriminate, which is the part worth checking:

* `test_bind_allocates_the_physical_memory` reads the runtime's device-memory gauge (`_rt_prof_memory`) around the call. No torch op can stand in for it — an op either materializes the allocation itself, or, for `zero_`, marks it logically zero without allocating anything — so this is the only witness the postcondition has from Python. Verified by removing the `rbln_set_raw_memory_alloc` call and rebuilding: every other test in the class still passed and this one failed.
* `test_is_huge_page_aligned`: a plain `torch.empty` of the same size is not 2 MiB aligned, checked directly.

The collective path has no new test — both sites make the same runtime call as before, through the shared helper. `test/distributed/test_process_group.py` covers them and now passes in full (406 tests, 4 NPUs).

## Notes

Not verified:

* **Nothing asserts actual huge-page backing.** Whether `madvise` is honoured depends on the host's THP setting, so the test asserts the 2 MiB alignment — which the provider does guarantee, and which is what keeps the copy path off its bounce buffer.
* `test_returns_zeroed_uint8_tensor`'s zero check is a contract assertion, not a discriminating one — freshly mapped pages read as zero anyway, so a wrapper that stopped zero-filling would only fail it by luck. The prefaulting behind it is not observable from Python.
* **No measurement accompanies either addition.** The mechanisms are read out of the runtime source (`multi_copy_cmd_buffer.cc` for the bounce, #12443 for the prefault), not measured.

### DIFF
--- a/c10/rbln/RBLNFunctions.cpp
+++ b/c10/rbln/RBLNFunctions.cpp
@@ -614,6 +614,10 @@ void free_nothrow(void* data) noexcept {
 void bind_device_memory(void* rbln_data, size_t nbytes) {
   RBLN_CHECK(rbln_data != nullptr, "bind_device_memory: rbln_data is nullptr");
   RBLN_CHECK(nbytes > 0, "bind_device_memory: nbytes must be positive, but got {}", nbytes);
+  // Unlike the other vmem-configuring leaves here, this one is reachable from a public Python
+  // entry point at any time, teardown included, where the raw rbln_* call would SEGFAULT
+  // rather than raise.
+  require_runtime("bind device memory");
   const auto vaddr = reinterpret_cast<uint64_t>(rbln_data);
   RBLN_LOG_DEBUG("bind_device_memory: vaddr={:#x} nbytes={}", vaddr, nbytes);
   RBLN_CHECK(
--- a/c10/rbln/RBLNFunctions.cpp
+++ b/c10/rbln/RBLNFunctions.cpp
@@ -614,9 +614,9 @@ void free_nothrow(void* data) noexcept {
 void bind_device_memory(void* rbln_data, size_t nbytes) {
   RBLN_CHECK(rbln_data != nullptr, "bind_device_memory: rbln_data is nullptr");
   RBLN_CHECK(nbytes > 0, "bind_device_memory: nbytes must be positive, but got {}", nbytes);
-  // Unlike the other vmem-configuring leaves here, this one is reachable from a public Python
-  // entry point at any time, teardown included, where the raw rbln_* call would SEGFAULT
-  // rather than raise.
+  // Reachable from a public Python entry point at any time, teardown included, where the raw
+  // rbln_* call would SEGFAULT rather than raise. The vmem-configuring leaves that only run
+  // downstream of an allocation do not repeat the check: malloc already made it.
   require_runtime("bind device memory");
   const auto vaddr = reinterpret_cast<uint64_t>(rbln_data);
   RBLN_LOG_DEBUG("bind_device_memory: vaddr={:#x} nbytes={}", vaddr, nbytes);
@@ -631,6 +631,9 @@ void bind_device_memory(void* rbln_data, size_t nbytes) {
 void set_device_layout_like(void* target_data, const void* ref_data) {
   RBLN_CHECK(target_data != nullptr, "set_device_layout_like: target is nullptr");
   RBLN_CHECK(ref_data != nullptr, "set_device_layout_like: ref is nullptr");
+  // Same reason as bind_device_memory: a public Python entry point, so nothing
+  // upstream has established that the runtime is loaded.
+  require_runtime("set the device layout");
   const auto target_vaddr = reinterpret_cast<uint64_t>(target_data);
   const auto ref_vaddr = reinterpret_cast<uint64_t>(ref_data);
   RBLN_LOG_DEBUG("set_device_layout_like: target={:#x} ref={:#x}", target_vaddr, ref_vaddr);
--- a/c10/rbln/RBLNFunctions.cpp
+++ b/c10/rbln/RBLNFunctions.cpp
@@ -611,6 +611,19 @@ void free_nothrow(void* data) noexcept {
   }
 }
 
+void bind_device_memory(void* rbln_data, size_t nbytes) {
+  RBLN_CHECK(rbln_data != nullptr, "bind_device_memory: rbln_data is nullptr");
+  RBLN_CHECK(nbytes > 0, "bind_device_memory: nbytes must be positive, but got {}", nbytes);
+  const auto vaddr = reinterpret_cast<uint64_t>(rbln_data);
+  RBLN_LOG_DEBUG("bind_device_memory: vaddr={:#x} nbytes={}", vaddr, nbytes);
+  RBLN_CHECK(
+      !::rbln::rbln_set_raw_memory_alloc(vaddr, static_cast<uint64_t>(nbytes)),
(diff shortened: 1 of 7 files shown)


torch_rbln/csrc/rbln/Module.cpp
```diff
@@ -212,6 +214,16 @@ void register_internal_api(py::module_& module) {
       },
       "Internal: mark RBLN virtual memory as zero-initialized (no host alloc)");
 
+  // Materialise a tensor's device allocation up front, for a consumer that reads the physical
+  // buffers out of band. Used by torch_rbln.bind_device_memory().
+  module.def(
+      "_bind_device_memory",
+      [](const at::Tensor& tensor) {
+        check_base_rbln_tensor(tensor, "bind_device_memory", "tensor");
```
rebel-jonghewk, line 222: what happens when we don't check_base_bln_tensor here?

c10/rbln/RBLNFunctions.cpp
```diff
@@ -611,6 +611,23 @@ void free_nothrow(void* data) noexcept {
   }
 }
 
+void bind_device_memory(void* rbln_data, size_t nbytes) {
+  RBLN_CHECK(rbln_data != nullptr, "bind_device_memory: rbln_data is nullptr");
+  RBLN_CHECK(nbytes > 0, "bind_device_memory: nbytes must be positive, but got {}", nbytes);
+  // Unlike the other vmem-configuring leaves here, this one is reachable from a public Python
+  // entry point at any time, teardown included, where the raw rbln_* call would SEGFAULT
+  // rather than raise.
+  require_runtime("bind device memory");
```
rebel-jonghewk, line 620: I can see that some do `require_runtime` but some don't . e.g (  set_device_layout_like, memcpy_h2v/v2h/..). Is there a rule for this?

torch_rbln/csrc/rbln/Module.cpp
rebel-chanheo: Two of the four clauses buy a readable error; the other two are load-bearing.

The lambda passes `tensor.data_ptr()` and `tensor.storage().nbytes()`, and `rbln_set_raw_memory_alloc` ends in `SetDeviceAllocConfiguration`, which looks the address up as a key and configures one flat region of that size starting there.

| clause dropped | slips through | result |
| --- | --- | --- |
| `is_privateuseone()` | a CPU tensor | a host pointer read as a vaddr → the runtime's `Invalid key_vaddr=...` instead of "must be an RBLN tensor" |
| `storage_offset() == 0` | `base[64:]` | an interior address is not a key → same runtime error |
| `is_contiguous()` | `base.t()` | **succeeds**: `data_ptr()` is still the base, so the whole allocation is bound flat — a call the API cannot honor for the strided view it named |
| `nbytes == numel * itemsize` | `base[:512]` of a 1024-element base | **succeeds**, binding the whole base rather than the 512 named |

The last two matter because a consumer DMAs the bound region out of band: "you bound something other than what you asked for" has to be an error, not a surprise. A whole-storage view (`base.view(...)`, `base[:]`) is still accepted, since it carries the allocation's own address and size.

c10/rbln/RBLNFunctions.cpp
rebel-chanheo: There is a rule, and this was a place it wasn't applied — fixed in 2742d76.

The rule is in malloc's comment: allocation is the gateway. require_runtime makes a raw rbln_* call on a torn-down runtime throw instead of SEGFAULT, and memcpy_*, streams, events and stats are only reachable because an allocation already succeeded. A public Python entry point isn't — which is why bind_device_memory has it, and set_device_layout_like should have had it too. It does now.

Cost is an atomic load plus a bool read behind a PLT call; through ctypes it measures at the harness's own floor, against a rbln_set_raw_memory_alloc several times that and an allreduce four orders of magnitude above both. ensureRcclRawMemory already made that call on every collectivebefore this PR.

rebel-jonghewk (approved):

PR #214, opened 2026-08-26
refactor(warm-cache): drive the rebel runtime by pybind method, not by mangled symbol

rebel-chanheo:
## Summary of Changes

The warm-cache hit path drove rebel's sync runtime by linking its C++ symbols — `rbln::PyRblnSyncRuntime::{PrepareInputs, PrepareOutputs, Run}`, declared in a forward-declaration header we maintained ourselves — and reached the instance by reading the C++ pointer out of the pybind11 object layout.

Neither is a contract rebel maintains. Those signatures ship in no header, and a C++ symbol name carries its parameter types, so adding a parameter to one renames the symbol we linked. With `-Wl,-z,now` the extension then fails to load, so every workload breaks at import, including those that never reach this path. `rebel_compiler#13059` adds a parameter to `PrepareInputs`.

This calls the runtime's `prepare_inputs` / `prepare_outputs` / `run` instead, resolved to bound methods once at install so a hit pays a call and not an attribute lookup. That is the surface rebel keeps stable for its own Python callers (`rebel/python/rebel/sync_runtime.py`), so a C++ signature change no longer reaches torch-rbln, and the parameter defaults rebel declares for those callers now apply here too. Three private-ABI dependencies go away: the mangled symbols, the pybind11 instance-layout reader (`_pybind_instance_raw_ptr`), and the Python type gate that existed only to make the raw pointer safe.

What remains is a dependency on four names — `prepare_inputs`, `prepare_outputs`, `run`, `_runtime_handle` — and losing one of them is silent: results stay correct, the warm cache stops serving.

**Layer:** torch-rbln C++ (`torch_rbln/csrc/rbln/`) and Python (`torch_rbln/_internal/`). No rebel-compiler change, no pin move.

**Paths exercised:** eager. The dispatch shim is eager-only; the hit path is reached by the shimmed elementwise and comparison `.out` ops.

## Which side notices a contract break

rebel-compiler lands changes to this surface before torch-rbln follows, so the three kinds of change need to be told apart — and only one of them should hold rebel up.

| rebel change | rebel CI (`test/rbln`, pinned torch-rbln) | torch-rbln CI (`test/internal` + `test/rbln`) |
| --- | --- | --- |
| parameter gains a default (`#13059`) | green | green |
| parameter becomes required, or a method is renamed | **green** — results are unaffected | **red** — the hit-path test |
| parameter changes meaning or position | **red** — results are wrong | **red** |

The middle row is the point: rebel can land the interface change, and the signal lands on the side that has to follow it. That is why the new tests live under `test/internal/` — rebel's CI runs `test/rbln` only (`tests/scripts/task_rebel_ci_torch_rbln_pytest.sh`), so a hit-path assertion placed there would block a change that costs nothing but this optimization.

Three tests carry it:

* `TestWarmCacheHitPath` — repeated dispatch of one input profile keeps advancing the hit counter. This is the red in the middle row.
* `TestWarmCacheContractBreak` — a runtime that rejects the call still produces correct values and counts no hit. This is the green in the middle row, asserted rather than assumed.
* `TestWarmCacheHandleGate` — a handle missing any of the three methods is refused before an entry is cached.

## Related Issues / Tickets

* Resolves #
* Related to #

## Type of Change

## Affected Modules

- [x] Backend

1. `uv run --no-sync pytest test/internal/test_warm_cache_internals.py test/rbln/test_dispatch_shim_precheck.py test/rbln/test_warm_cache_view_on_device.py`
2. Verify the three warm-cache classes above pass.
3. Edge case: rewire `TestWarmCacheContractBreak` to pass the real handle through instead of the rejecting stand-in — the hit-counter assertion must then fail. That check is what shows the test can detect the case it covers.

## Verification

| Command | Result |
| --- | --- |
| `uv run --no-sync lintrunner -m origin/dev` | exit 0, no findings |
| `uv run --no-sync python test/run_tests.py --suite=core` | 4 lanes passed, 0 failed |
| `uv run --no-sync pytest test/internal/ test/rbln/test_dispatch_shim_precheck.py test/rbln/test_warm_cache_view_on_device.py` | passed |
| `uv run --no-sync pytest test/rbln/test_graph_eager_mode.py` | passed |

The built extension carries no undefined `rbln::PyRblnSyncRuntime` symbol; on `dev` it carries three.

## Notes

**Performance.** The hit path now crosses the pybind boundary three times per call, which costs a constant per hit.

Harness: the per-segment hit-path timers (`_dispatch_shim_warm_segments_dump`), 64-element fp16 `add.out` and `mul.out`, 15 blocks of 2000 calls across 3 processes per side, `TORCH_RBLN_DEPLOY=ON`, `RBLN_VERBOSE` and `TORCH_RBLN_EAGER_MALLOC` unset, torch 2.11.0, rebel-compiler 0.11.3.dev0. Both sides built with `uv pip install -e . --no-build-isolation`.

* the input-prepare segment grows about 2.4x and the output-prepare segment about 3.3x, consistently across processes;
* the `run` segment shows no resolvable change — its process-to-process spread is an order of magnitude larger than the added cost;
* the added cost is a few percent of the whole hit, which device execution dominates.

The cost is per shimmed op, so a workload with many eager ops per step multiplies it. That has not been measured, and it is what would justify moving these three calls to an entry point in rebel's shipped C API; on this evidence it does not.

**A test that does not fail without this change.** `TestWarmCacheHitPath` passes on `dev` too — the hit path works there. It guards the failure mode this change introduces, which an install-side assertion cannot see. The other two do fail without the change: `TestWarmCacheHandleGate` covers a function this PR adds, and `TestWarmCacheContractBreak` asserts a degradation path that did not exist while the symbols were linked.

**Released wheels.** v0.2.2 through v0.4.0rc0 link the mangled symbols and cannot follow this change. They break at import against a rebel-compiler carrying `#13059`, and so does rebel's own `atom-pytest-torch-rbln` step, which pins `TORCH_RBLN: "0.4.0rc0"`. Closing that window needs either a rebel-side overload kept until a release containing this PR exists, or that pin raised to it. Raising the pin does not move the hit-path test into rebel's lane: its command names `test/rbln`, so `test/internal` stays out of it whichever version is pinned, and the table above keeps holding.


rebel-daeyang (approved):
LGTM

PR #229, opened 2026-09-02
feat(memory): serve a strided device copy as one compiled program

rebel-chanheo:
## Summary of Changes

A device→device `copy_` whose source is a view has two routes: the strided v2v engine, which spends a descriptor per contiguous run, and above its cap a host round-trip. Neither fits a copy that is one reshape away from contiguous — a paged KV block read head-major and written token-major breaks into runs of a single row, far past the cap, so what it actually gets is the bounce.

This routes such a copy through the view machinery the compile path already has. `prepare_args_view_aware` classifies the chain, `get_view_op_module` replays it *inside* the compiled graph, and the device reads the base buffer and writes the destination in one program. Nothing in the new code knows what a permutation is: whatever the detector classifies — a slice, a reshape, a select, or any composition — this covers.

**`feat(memory): teach the view detector to merge and split dims`**

`_apply_view_step` has always known how to replay a `reshape`, but nothing ever produced one: the candidate generator offers permute, squeeze, unsqueeze, narrow, expand and select, and the simulator rejects any step it does not recognise. A chain that flattens before it permutes — `kv.view(rows, heads, block, dim).permute(0, 2, 1, 3)`, the shape a paged KV transfer writes — came back unclassified and fell through to `.contiguous()`.

A merge or a split is a pure view exactly when the layout is contiguous from offset 0, which is where the new shape's strides follow from its sizes. The simulator now checks that, and the generator proposes one only from such a state. Both families are bounded by the target's own shape, so the search frontier does not grow.

This stands on its own — every consumer of the detector picks it up, since an unclassified view is materialised on the host.

**`feat(memory): serve a strided device copy as one compiled program`**

`torch_rbln::copy_strided_view` — schema declared in `RBLNRegisterOps.cpp`, implementation in Python because it drives the compile path — plus one branch in `tensor_copy_from_rbln_to_rbln`, ahead of the strided walk. `copy_` reaches it through the dispatcher rather than a hand-installed function pointer.

The namespace is this backend's own, not `rbln_custom_ops`: that one belongs to rebel-compiler and its names mirror the aten op each converts, so a `view_copy` there collides the moment it grows a converter for `aten::view_copy`, and two defs of one schema abort at import.

Two conditions decide whether to reach for it, and both are facts about the compiler quoted at the call site rather than tuning knobs:

- **dtype** — of the two dtypes the compile path dispatches, only bf16 reaches the device as itself; f16 is rewritten to a narrower float. `copy_` is not a compute op, so the fast and slow routes have to return the same values.
- **last-dim alignment** — below the compiler's threshold the tensor stays on the host, where the program is far slower than the walk it would replace.

Whether the view itself can be replayed is not a condition here: the shared detector answers that, and `view_copy` returns false when it cannot. That check runs *before* `compile_and_run_view_aware`, because an unclassified view takes `.contiguous()` in there, which re-enters `copy_`.

Not covered, each with a test asserting the existing route still serves it: a strided *destination* (the program writes a contiguous buffer, and scattering into a view is not something the out-tensor path can express) and a copy that crosses devices (one program reads and writes one device).

---

## Related Issues / Tickets

* Related to rebellions-sw/torch-rbln-internal#38
* Related to rebellions-sw/fsw-inference#540

---

## Type of Change

---

## Affected Modules

- [x] Backend
- [x] Ops/Kernels
- [x] `torch.compile`
- [x] Memory

---

## How to Test

```bash
uv run --no-sync pytest test/rbln/test_view_copy.py
uv run --no-sync python test/run_tests.py --suite=core
```

`test_view_copy.py` asserts the **route**, not only the result — the strided walk and the host bounce are correct too, so a result alone cannot tell them apart. It takes two counters. Every shape in it is past the strided engine's cap, so the host-bounce counter separates served from declined. That is not sufficient on its own: a served copy can still reach the host, because below the compiler's last-dim alignment the program is lowered with its input on the host and the runtime feeds it there without recording a bounce — confirmed by calling the op directly on an unaligned shape, which reports no bounce and a `v2h`/`h2v` pair. So the served cases also assert the runtime ran no host primitive.

- six classifiable chains (permute, reshape-then-permute, split-then-permute, select-, narrow-, transpose-last-two) take the program, on the first call and on repeats
- f16 / f32 / i32 / bool, three unaligned last dims, a strided destination, an unclassifiable `as_strided`, a contiguous copy and a device→host copy all keep the existing route and return the right values
- a copy across two logical devices keeps the existing route and returns the right values
- a destination at a non-zero storage offset is served but through the compile path's own buffer, so the result is moved across with a device-to-device copy rather than written in place
- one test asserts the detector still classifies every chain the others rely on, so a detector regression names itself instead of showing up as a slowdown

Verified that the tests fail without the change: 12 of 26 fail with the `copy_` branch disabled, 5 with the detector's reshape candidates removed, the three unaligned cases with the alignment gate removed, and the cross-device case with the device-equality gate removed.

Regression run (copy / view / profiler / warm-cache / graph-eager / op-caching / registered-ops): **3101 passed, 7 skipped.** The 16 failures in `test_add_mixed_devices` are not from this change — every logical device above `rbln:0` returns wrong values for a plain `add`, on a build containing none of this; reported separately. `lintrunner -m origin/dev -a` clean.

---

## Notes

**Measured on the KV-transfer benchmark in rebellions-sw/fsw-inference#540** — paired runs, alternating branches, n=9 each, one logical device, `rebel-compiler 0.11.3.dev108` (inside the `pyproject.toml` range, above the `constraints-build-dev.txt` pin). The compiler carried the `DynamoRuntime.run(out=)` change from rebellions-sw/rebel_compiler#13290 because the bespoke arm does not run without it; this branch does not use that path — it reaches the compile path through `out_tensor_context`, and its `out` argument stays `None`, which is also why it runs on a stock compiler at all (verified: the KV transfer's two copies come back bit-exact with no host primitive). Against a bespoke compiled-permute implementation of the same idea, total throughput is indistinguishable — the mean paired difference is under half a percent, and run-to-run spread on this benchmark is several times that. Split by direction, this one is ahead on gather and behind on scatter by similar margins; the scatter side is reproducible and its cause is not identified. What this buys over the bespoke version is that it needs no compiled-permute module, no C++ hook, and no compiler-side change, and it generalises past permutations.

Against the routes it replaces, a classifiable strided copy that used to round-trip the host is now two orders of magnitude cheaper.

The compile path is reached through `out_tensor_context`, so this needs nothing from `DynamoRuntime.run(out=)` — verified against a stock compiler build.

### DIFF
--- a/aten/src/ATen/native/RBLNRegisterOps.cpp
+++ b/aten/src/ATen/native/RBLNRegisterOps.cpp
@@ -130,6 +130,16 @@ TORCH_LIBRARY_IMPL(_, AutogradPrivateUse1, m) {
   m.fallback(torch::CppFunction::makeFromBoxedFunction<&autograd_fallback>());
 }
 
+// Operators this backend owns. `rbln_custom_ops` belongs to rebel-compiler, whose names
+// mirror the aten op each one converts -- `view_copy` is taken there the moment it adds a
+// converter for `aten::view_copy`, and two defs of one schema abort at import.
+TORCH_LIBRARY(torch_rbln, m) {
+  // `copy_` reaches this from C++ for a device->device copy whose source is a classifiable
+  // view; the implementation is Python because it drives the compile path. Answers false
+  // when the view cannot be replayed, leaving the caller on its existing route.
+  m.def("copy_strided_view(Tensor src, Tensor(a!) out) -> bool");
+}
+
 // ATen operations registration for the RBLN backend (PrivateUse1)
 TORCH_LIBRARY_IMPL(aten, PrivateUse1, m) {
   // Operations that use the device runtime API
--- a/aten/src/ATen/native/rbln/RBLNCopy.cpp
+++ b/aten/src/ATen/native/rbln/RBLNCopy.cpp
@@ -16,6 +16,7 @@
 
 #include <c10/rbln/RBLNProfiler.h>
 #include <rebel/runtime/api/rbln_runtime_api.h>
+#include <torch/library.h>
 
 #include <cstddef>
 #include <optional>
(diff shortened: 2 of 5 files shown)


rebel-jinhwan (approved):
LGTM

PR #231, opened 2026-09-02
test(contract): judge the remap verdict against the runtime's pool, not /dev nodes

rebel-chanheo:
## Summary of Changes

`test_cotenant_availability_probe_leaves_the_mapping_remappable` fails in rebel-compiler CI
(`[OB] ATOM torch-rbln Tests`, since the `TORCH_RBLN` pin moved to 0.4.0) with
`AssertionError: 'frozen' != 'applied'`. Nothing is frozen there.

The verdict is measured by remapping `RBLN_DEVICES` and comparing the count the runtime
reports. With one usable NPU a live mapping and a frozen one report the same number, so the
measurement needs a "cannot tell" gate — and that gate counted `/dev/rbln*`. Device nodes are
not the number of NPUs the runtime can use: a step asking for `npu: count: 1` on a multi-NPU
host passes the guard, and its legitimate 1-and-1 reads as a freeze.

Test-only, four commits:

1. **Judge against the runtime, not device nodes.** The probe reports the two raw counts;
   `Probe.remap` draws the verdict against a pool scanned in a process of its own. That
   process is the point: the runtime's `RBLN_DEVICES` seal is process-local, so an in-process
   query answers with the sealed value once anything here has used a device —
   `physical_device_count()` included, since it is the same `rbln_get_device_count()` the
   probe measures — and the state under test would be deciding its own gate.
2. **Measure in a forked child.** Reading the counts rewrites `RBLN_DEVICES`, and the probe
   body may have left threads running (`DataLoader(pin_memory=True)` keeps a pin-memory
   thread). setenv/unsetenv beside another thread's getenv is a data race, and
   `c10/rbln/DeviceMappingManager.h` rules out mutating `RBLN_*` alongside a query at all. The
   seal is inherited across fork, so the child reads the parent's latch state while its own
   edits stay private to it.
3. **Name devices from this job's own pool.** The scenarios and three probe bodies hard-coded
   id 0. Where a job was allocated other NPUs that is a co-tenant's device — measured wrong,
   and on the allocating paths claimed.
4. **A failed measurement costs the counts, not the record.**

## Related Issues / Tickets

* No issue filed yet — surfaced by rebel-compiler CI after rebellions-sw/rebel_compiler#13301
  bumped the `TORCH_RBLN` pin to 0.4.0.

## Type of Change

- [x] `other` — test-only (contract test harness), plus one `docs/TEST_GUIDE.md` convention

## Affected Modules

- [x] Docs

## How to Test

1. `uv run --no-sync pytest test/rbln/test_privateuse1_contract.py` on 4x RBLN-CA25 →
   **28 passed, 3 xfailed**, same verdicts as before the change.
2. One-NPU allocation, the shape that fails in CI:
   `RBLN_DEVICES=2 pytest test/rbln/test_privateuse1_contract.py -k "<the five remap clauses>"`
   → **4 passed, 1 skipped**; scenarios move to device 2, the verdict is still drawn
   (`pool=1/4`), and only the two-NPU clause skips, on the count it reads back.
3. Verdict function, all branches: pool 4 with counts `[1,4] / [1,1] / [2,2] / [0,4] / None`
   → `applied / frozen / frozen / None / None`; pool 1 and pool 0 with `[1,1]` → `None`
   (previously a `frozen` false positive).
4. The measurement no longer touches the probe's environment: the pin-memory probe now ends
   with `RBLN_DEVICES` as the scenario set it, where before it ended unset.
5. `pytest -n 4` → 5 passed. `lintrunner -m origin/dev` → no issues.

## Notes

* The gate is the **host** pool while the ids are the **runner's selection** — split on
  purpose. A probe runs with the selection stripped and its second count unsets
  `RBLN_DEVICES`, so the host is what it can enumerate; narrowing the gate as well would skip
  the clause wherever a runner happens to select one NPU.
* With one usable NPU the clause cannot be verified at all — a real freeze reads the same as a
  live mapping. That is now an honest skip rather than a false failure, and the skip/failure
  line carries `pool=N/M`, so the next compiler-CI run says which case it is: one NPU in the
  container (environment), or two-plus and still frozen (a runtime bug worth chasing).
* No production code changes. One extra subprocess per pytest worker (~0.1s scan, cached).

## Checklist

* [ ] This PR is linked to a corresponding issue — none filed; happy to open one in
  `rebellions-sw/fsw-inference` if you want it tracked there
* [x] Linting passes
* [x] All CI tests pass
* [x] Relevant documentation has been updated (`docs/TEST_GUIDE.md`)

### DIFF
--- a/docs/TEST_GUIDE.md
+++ b/docs/TEST_GUIDE.md
@@ -372,6 +372,11 @@ Conventions specific to it:
 - **A clause not satisfied yet is a `strict=True` xfail** naming the work that closes it,
   rather than being omitted. `xfail_strict = true` is the project default, so an unexpected
   pass fails the suite and signals that the marker should be removed.
+- **The remap verdict needs a reference measurement.** With one usable NPU a live mapping and
+  a frozen one report the same device count, so `Probe.remap` is conclusive only when a pool
+  scanned in a clean process shows two or more; `/dev/rbln*` is not that count, since a
+  container can show more device nodes than the runtime can use. Scenarios name devices from
+  this job's own selection instead of assuming id 0 is theirs.
 
 ---
 
--- a/test/rbln/test_privateuse1_contract.py
+++ b/test/rbln/test_privateuse1_contract.py
@@ -182,11 +182,12 @@ print(json.dumps({"ids": [i for i in range(127) if npu_is_available(i)]}))
 def _host_pool() -> tuple:
     """Every physical NPU id on this host, scanned in a process of its own.
 
-    The reference a remap verdict is judged against, and it has to be another process. The runtime's RBLN_DEVICES seal is process-local, so once
-    anything here has used a device an in-process query answers with the sealed value --
-    ``physical_device_count()`` included, since it is the same ``rbln_get_device_count()`` the
-    probe measures -- and the state under test would be deciding its own gate. Stripping the
-    selection variables is what makes the answer the host rather than a selection of it.
+    The reference a remap verdict is judged against, and it has to be another process. The
+    runtime's RBLN_DEVICES seal is process-local, so once anything here has used a device an
+    in-process query answers with the sealed value -- ``physical_device_count()`` included,
+    since it is the same ``rbln_get_device_count()`` the probe measures -- and the state under
+    test would be deciding its own gate. Stripping the selection variables is what makes the
+    answer the host rather than a selection of it.
 
     Empty when the runtime, the driver or the NPUs are absent, and when the scan itself fails:
     callers then treat every remap verdict as undetermined.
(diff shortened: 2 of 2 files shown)


rebel-huijong:
Just to confirm my understanding: if we ask the runtime for the device, it gets sealed within that process anyway, so we were originally counting /dev/devices* instead. But there was some issue with that in the CI environment, so this was changed to fork a separate process and count devices through the runtime inside that process. Is that right?

rebel-huijong (approved):

rebel-chanheo:
> this was changed to fork a separate process and count devices through the runtime inside that process. Is that right?

Yes, that's it. One nuance: the baseline scan is a fresh subprocess, not a fork -- a fork would inherit the seal. The fork is only for the measurement itself.

PR #233, opened 2026-09-03
build: extend the supported Python range to 3.14

rebel-chanheo:
## Summary of Changes

Extends the supported Python range to 3.14.

The pin moves to `0.11.3.dev185+g9480a7b8.prod`, which publishes cp310 through cp314. The build it replaces (`dev160+g0ebb3f84.prod`, from #230) publishes cp312 alone, narrower than the matrix the release lanes fan out to, so this pin restores that as well as adding 3.14.

torch needs nothing: `torch==2.11.0+cpu` already publishes cp314 wheels, and Dynamo refuses only 3.15 and above. The rest of the first commit is `requires-python`, the release lint matrix, the `workflow_dispatch` input descriptions, and the two places the docs state the range.

The second commit is what building on 3.14 turned up. The build frontend hands CMake the interpreter it is building for, but the CMake modules look it up with `find_package(Python3)`, which reads `Python3_EXECUTABLE` and ignores the `Python_EXECUTABLE` that was being passed. On a host with more than one interpreter, CMake resolved Python, torch, and pybind11 from whichever one PATH led with: the wheel came out tagged `cp314` carrying an `_C.cpython-312-*.so`, and importing it landed on the `_C/` stub directory and failed with `no attribute '_dispatch_dtypes'`. Stub generation had the same hole — `find_program(STUBGEN stubgen)` — and a `stubgen` belonging to another interpreter cannot import the extension it is asked to describe, so `_C.pyi` was never written and `cmake --install` failed on the missing file. CI never saw either one because its containers hold exactly one interpreter.

The third commit moves two lint pins. `pandas 2.2.x` and `optree 0.13.x` publish no cp314 wheel, so a 3.14 lint lane resolves them to sdists and compiles both from source on every cold cache — they do install, which is why it is easy to miss. `pandas 2.3.3` and `optree 0.17.0` are the first releases with cp314 wheels and both still cover 3.10 through 3.13. `mypy` needed nothing: 1.13.0 runs on 3.14 and reports no errors in `torch_rbln/`.

A fourth commit moves the build image. rblness builds py3.14 through an opt-in path (rebellions-sw/rblness#198), and a release -- the only thing that promotes an image to GHCR -- cannot be cut right now, so `ghcr.io/rebellions-sw/rbln-build:0.11.0-ubi9` will never gain a py3.14 tag. The draft registry has one that carries every Python from 3.10 to 3.14, so the three workflows that run a job per Python (wheel build, source lint, nightly torch) move there together rather than 3.14 alone. `HARBOR_USERNAME` and `HARBOR_PASSWORD` are already set on the repository. `rbln-test` stays on GHCR -- those jobs pin py3.12.

---

## Related Issues / Tickets

* No linked issue.

---

## Type of Change

- [x] `other` — supported-version range, plus a build fix it exposed

---

## Affected Modules

- [x] Build/Tools
- [x] Docs

---

On a host that also has other interpreters installed, against Python 3.14.5:

1. `uv venv --python 3.14 .venv-314 && uv pip install --python .venv-314/bin/python -c constraints-build-dev.txt rebel-compiler "torch==2.11.0+cpu"`
2. `.venv-314/bin/python -m hatchling build -t wheel` — expect `torch_rbln-*-cp314-cp314-manylinux_*.whl` holding `_C.cpython-314-x86_64-linux-gnu.so`
3. Install it and, from outside the source tree, import and run one device op.
4. `pytest test/rbln/test_abi_check.py test/rbln/test_device_mapping.py test/rbln/test_copy_v2v.py`
5. `uv sync --python 3.14 --no-install-project --no-install-package rebel-compiler`, then `mypy --config-file mypy.ini torch_rbln`

Expected: the wheel's ABI tag and its `_C` suffix agree, the op returns on device, the tests pass, and the lint environment resolves to wheels rather than sdists.

---

```
-- Installing: torch_rbln/_C.cpython-314-x86_64-linux-gnu.so
dist/torch_rbln-0.4.1.dev3+g46bb45ab9-cp314-cp314-manylinux_2_35_x86_64.whl

VERIFY python/torch/rebel/torch_rbln: 2.11.0+cpu 0.11.3.dev185+g9480a7b8 0.4.1.dev3+g46bb45ab9
VERIFY device op: [2.0, 2.0, 2.0, 2.0]

261 passed, 17 skipped in 263.62s
```

Before the second commit, the same build produced a `cp314`-tagged wheel containing `_C.cpython-312-x86_64-linux-gnu.so`, and importing it raised `AttributeError: module 'torch_rbln._C' has no attribute '_dispatch_dtypes'`.

---

## Checklist

---

## Notes

Stacked on #230 — that is the base branch, and the pin here supersedes the one it introduces. Merge that first.

Two things this PR cannot settle on its own:

* The release lint lane now includes 3.14. It pulls the draft-registry image this PR switches to, which does have a py3.14 tag, but only a PR targeting `main` runs that lane -- nothing here exercises it.
* `pypi.rbln.ai` serves no cp314 wheel for any released rebel-compiler — the cp314 builds are on the internal index only. Until a release carries them, the range this PR documents holds for the internal lanes but not for a source build off the public index.

#230 pins its bfloat16 canary to rebel's `dev` at `g0ebb3f84`; this pin is 25 commits later on the same line, so if the ATOM promotion was fixed in between, the strict `xfail` there flips to an unexpected pass.


rebel-huijong (approved):

rebel-clee (approved):
LGTM

PR #235, opened 2026-09-03
feat(memory): add torch.rbln.physical_layout()

rebel-jonghewk:
### Summary

Adds `torch.rbln.physical_layout(tensor)`: where an RBLN tensor's bytes physically sit on the device, one `PhysicalShard(node_id, chiplet_id, nbytes, device_addr, shape)` per device area, plus the logical shape/dtype and the physical shape/dtype/itemsize of the allocation's transform. It is the input for per-chiplet KV cache sizing (`nbytes_per_chiplet()` divided by the compile-time block hint gives bytes per block per chiplet) and for placement diagnostics (a 4-way head split reports `(4, 4, 1, 128, 2048)` per shard; a 4-way replication reports the full shape four times).

Backed by the librbln C ABI `rbln_get_memory_info()` — `MemoryInfo.device_areas` added in rebellions-sw/rebel_compiler#13405 (ABI 4) — through a new `_C._get_memory_info` binding. No `rebel._C` import on the torch side.

### Details

- **`_C._get_memory_info(tensor)`** (`csrc/rbln/Module.cpp`) — validates the tensor with the same rule as `bind_device_memory` (RBLN tensor covering its whole storage), calls `c10::rbln::get_memory_info(data_ptr())` (`require_runtime` guard), and returns `physical_shape`, `physical_dtype` and `device_areas` as one dict per area keyed like `PhysicalShard`, so Python builds `PhysicalShard(**area)`. `physical_dtype` is a real `torch.dtype` via `getTHPDtype` from the new `c10::rbln::to_scalar_type()` (inverse of `to_rbln_dtype()`), or `None` for a dtype torch lacks — DLFloat16, the usual fp16/bf16 on-device storage.

- **`torch.rbln.physical_layout()`** (`memory.py`) — wraps the dict into `PhysicalLayout(logical_shape, logical_dtype, physical_shape, physical_dtype, physical_itemsize, shards)`. `physical_itemsize` is derived from an area (`nbytes / prod(shape)`), so it is set even when `physical_dtype` is `None`. A tensor whose allocation is not bound yet (no torch op or graph has touched it, no `bind_device_memory`) raises `no device memory yet`. A flat 1:1 binding has no transform: every `physical_*` field and each shard's `shape` is `None`.

- **Timing** — the physical view exists only after the allocation is bound, so the KV sizing flow reads the layout after warm-up and before releasing the tensor. A dynamic (`mark_dynamic`) dim is reported as the concrete value bound at that point, since the runtime regenerates the transform with resolved sizes.

- **Tests** (`test/rbln/test_physical_layout.py`) — API visible and CPU tensor rejected as plain tests; the device tests go through `instantiate_device_type_tests(only_for="privateuse1")`: bound flat buffer is one shard with `physical_* is None`; compiled-op output covers its bytes and every shard satisfies `prod(shape) × physical_itemsize == nbytes`; unbound tensor and interior view raise. `test_cpp_RBLNFunctionsTest.ToScalarTypeInvertsToRblnDtype` pins the dtype inverse.

### Next Steps

- vllm-rbln: replace the per-artifact `kv_cache_memory_profile()` merge with `physical_layout(kv).nbytes_per_chiplet()` after warm-up (dynamic KV proposal).
- Per-shard `slice_axes` (axis/begin/end) is not exposed; the sliced shape covers sizing and placement. Add it next to `shape` if the KV connector needs it.
- `lintrunner` was not run locally (init download timed out); `ruff check`/`ruff format --check` and `clang-format` pass on the changed files.
- `c10::rbln::get_memory_info()` now logs its cost at debug instead of warn, since `physical_layout()` calls it on purpose; the header note still tells hot-path callers to prefer `get_torch_device_id()`.

---

## Related Issues / Tickets

* Related to rebellions-sw/rebel_compiler#13405 (required: `MemoryInfo.device_areas`, `RBLN_ABI_CURRENT` 4)

---

## Type of Change

---

## Affected Modules

- [x] Memory

---

1. Build against a rebel_compiler that includes rebellions-sw/rebel_compiler#13405.
2. Run `pytest test/rbln/test_physical_layout.py` on an RBLN device.
3. Expected: 6 passed. On REBEL a compiled matmul output reports one shard per chiplet it was placed on, each with `prod(shape) × physical_itemsize == nbytes`.

---

## Checklist

---

## Notes

CI compiles against the pinned `rebel-compiler` wheel, which does not yet have `MemoryInfo.device_areas`; the C++ build will fail until rebellions-sw/rebel_compiler#13405 lands and the pin moves. Kept as draft until then. All six tests pass on the dev box (REBEL); `test_set_device_layout_like.py` and `test_tensor_memory.py` were run alongside (52 passed in total) to check the neighbouring memory APIs.

### DIFF
--- a/c10/rbln/RBLNFunctions.cpp
+++ b/c10/rbln/RBLNFunctions.cpp
@@ -352,13 +352,52 @@ c10::DeviceIndex get_torch_device_id(const void* data) {
   return static_cast<c10::DeviceIndex>(torch_device_id);
 }
 
+std::optional<c10::ScalarType> to_scalar_type(::rbln::DataType rbln_dtype) {
+  switch (rbln_dtype) {
+    case ::rbln::DataType::Bool:
+      return c10::kBool;
+    case ::rbln::DataType::UInt8:
+      return c10::kByte;
+    case ::rbln::DataType::Int8:
+      return c10::kChar;
+    case ::rbln::DataType::Int16:
+      return c10::kShort;
+    case ::rbln::DataType::Int32:
+      return c10::kInt;
+    case ::rbln::DataType::Int64:
+      return c10::kLong;
+    case ::rbln::DataType::Float16:
+      return c10::kHalf;
+    case ::rbln::DataType::Float32:
+      return c10::kFloat;
+    case ::rbln::DataType::Float64:
+      return c10::kDouble;
+    case ::rbln::DataType::Float8_e4m3:
+      return c10::kFloat8_e4m3fn;
+    case ::rbln::DataType::Float8_e5m2:
+      return c10::kFloat8_e5m2;
+    case ::rbln::DataType::BFloat16:
+      return c10::kBFloat16;
+    case ::rbln::DataType::Complex32:
+      return c10::kComplexHalf;
+    case ::rbln::DataType::Complex64:
+      return c10::kComplexFloat;
+    case ::rbln::DataType::Complex128:
+      return c10::kComplexDouble;
+    case ::rbln::DataType::Undefined:
+    case ::rbln::DataType::CustomFloat16:
+      return std::nullopt;
+  }
+  return std::nullopt;
+}
+
 ::rbln::MemoryInfo get_memory_info(const void* data) {
   RBLN_LOG_DEBUG("data={}", fmt::ptr(data));
   RBLN_CHECK(data != nullptr, "data cannot be nullptr");
   // get_memory_info() does a full VMemory JSON serialize+parse round-trip on
-  // every call. Warn so unexpected hot-path callers are easy to spot in logs;
-  // when only the device id is needed, get_torch_device_id() is the cheap path.
-  RBLN_LOG_WARN(
+  // every call; when only the device id is needed, get_torch_device_id() is the
+  // cheap path. Debug (not warn): torch.rbln.physical_layout() calls this on purpose.
+  RBLN_LOG_DEBUG(
       "get_memory_info({}) performs a full VMemory JSON round-trip (slow) — avoid on performance hot paths; "
       "use get_torch_device_id() when only the device id is needed",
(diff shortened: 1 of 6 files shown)


rebel-jonghewk:
Addressed in the second commit.

1. **Field order** — the tuple is now `(node_id, chiplet_id, size, device_addr, shape)`, matching `PhysicalShard`. Confirmed the swap by reading both sides; the device tests that would have caught it are still pending on hardware (see below).
2. **physical_dtype string** — added `PhysicalLayout.physical_itemsize` from a runtime-dtype size table (`CustomFloat16` → 2). The test uses it instead of `torch.empty(dtype=...)`. `physical_dtype` stays a `torch.dtype | str`.
3. **Test convention** — `test_api_visible` and `test_cpu_tensor_raises` are module-level (device-free); the four device tests are in `TestPhysicalLayoutOnDevice` under `@requires_logical_devices(1)`.
4. **c10 wrapper** — the binding calls `c10::rbln::get_memory_info(tensor.data_ptr())`, so it gets `require_runtime()` and the cost warning.
5. **Cost** — docstring now says it is a diagnostic call (full serialization round-trip), read once after warm-up.
6. **2 MiB** — it is `kDeviceAllocAlignment = 2ull * 1024 * 1024` in `rebel/src/pyrbln_impl/compiled_model.cc` ("Device DRAM allocation granularity ... see GetAlignedSize in buffer_utils"), so the number stays.

Device runs are still blocked by another job holding all 8 NPUs on the dev box; a monitor runs the file as soon as they free up.

rebel-jonghewk:
Device run done (REBEL, latest commit): `test/rbln/test_physical_layout.py` 6/6, and 52/52 together with `test_set_device_layout_like.py` + `test_tensor_memory.py`. The PR body's pending note is updated.

PR #238, opened 2026-09-05
refactor(eager): bind out= through DynamoRuntime.run instead of rebel's out-tensor stash

rebel-chanheo:
## Summary of Changes

Stacked on #236 (the base of this PR is `chan/warmcache-wrapped-scalar`); the two commits of rebellions-sw/fsw-inference#232 are included and this PR supersedes it. After #236 lands the base moves to `dev`.

torch-rbln bound a caller's `out=` buffer for the compile miss path by setting `rebel.core.torch_eager.eager_execution_helper()`, a process-global list the runtime reads while filling its outputs. Two threads dispatching at once could hand one thread's result to the other's buffer (rebellions-sw/torch-rbln-internal#37). This PR removes every use of that stash from torch-rbln.

**What replaces it.** The Dynamo-compiled callable has no slot for `out`, and a device copy of the runtime's output into `out` is not a substitute: when `out` is host-latest (filled from the CPU, or by a CPU fallback) the runtime serves that copy on the host view and `out` stays off the device until something else consumes it. So the miss path binds `out` the way the C++ warm hit already does, as the program's output buffer, through `DynamoRuntime.run(out=)`:

* On the first call for a compile-cache key, `compiled()` runs once (that is where the rebel backend builds the runtime, into a buffer of its own that is dropped) and a second `run(out=[out])` binds the caller's buffer. This costs one extra run, once, under the compile that just happened.
* On every later Python-path call (view inputs, shim bail-outs), the cached runtime is driven directly with `run(*tensors, out=[out])`; the compiled callable and its Dynamo guards are skipped. `compile_rbln_cached_entry` returns a `CompiledOp` holding the callable and the runtime holder the backend filled.

**What skipping the guards requires.** The compile-cache key has to separate everything the callable would have recompiled for: device, shapes, dtypes, scalar arguments (#236), and the argument identity pattern, since Dynamo folds a tensor passed twice into one graph input. Grad mode and `requires_grad` were checked on the device and do not recompile this callable, so they are not in the key (keying grad mode would compile every op once more under `no_grad`). The graph's input order is read back from the runtime after the first run rather than assumed: a view replay puts its base ahead of the other arguments, and a 0-dim CPU constant is no input at all.

**Also removed.** `device/context_holder.py` (the stash wrapper), `kernels/custom_transpose.py` (no caller), and the empty stash contexts around the sdpa kernels.

**Layer.** torch-rbln Python, eager path. Nothing changes in the C++ shim. The build pin moves to the rebel-compiler that accepts device tensors in `run(out=)` (rebellions-sw/rebel_compiler#13376).

---

**Update (pin moved to `0.11.3.dev238+gff063162`).** That build carries rebellions-sw/rebel_compiler#13453, which lets the runtime skip the host->device push when a program output overwrites its whole vmemory entry. With it, binding a host-latest or `torch.zeros` `out` moves no bytes to the device on either path. `test_binding_out_moves_no_bytes_to_the_device` pins that through the runtime's real-transfer counter on the C++ warm hit and on the view-input Python path (the compile miss is left out: the compile itself uploads the program). On the previous build (dev235) the same test fails with the tensor's size pushed per bind. dev238 is the CI build of the #13453 branch merged onto `dev`; the pin should follow the post-merge build before this lands.

---

## Related Issues / Tickets

* Related to rebellions-sw/torch-rbln-internal#37
* Supersedes rebellions-sw/fsw-inference#232
* Depends on #236 (stacked)

---

## Type of Change

---

## Affected Modules

- [x] Backend
- [x] Ops/Kernels
- [x] `torch.compile`
- [x] Build/Tools

---

Needs `rebel-compiler==0.11.3.dev238+gff063162` (the pin); it is a cp312 wheel, so a Python 3.12 environment.

1. `uv pip install -e . --no-build-isolation` (codegen changes; the generated shim must stop importing `custom_transpose`).
2. `uv run --no-sync pytest test/rbln/test_internal_op_utils.py -k TestOutTensors -x`
   Expected: pass. Without the change, `test_out_is_filled_without_the_rebel_global_out_stash` fails (`set_out_tensor` called once), `test_host_latest_out_becomes_device_resident_on_every_path` fails (a CPU-filled `out` stays host-latest after the miss-path copy), and `test_compile_miss_with_out_keeps_no_extra_output_alive` fails.
3. `uv run --no-sync python test/run_tests.py --suite=core`
4. `uv run --no-sync lintrunner -m origin/dev`

Run here on RBLN-CR03 (`RBLN_DEVICES=4,5,6,7`, NPUs 0-3 were held by another user's job):

* `test_internal_op_utils.py test_tensor_memory.py test_llama_ops.py test_accelerator_contract.py test_eager_async_buffer_regression.py` → 226 passed.
* core suite: parallel pass 4281 passed / 29 skipped / 3 xfailed; single-worker pass 12 passed / 4 xfailed. The 8 failures across both passes are the subprocess and probe tests that open physical NPU 0 (`test_dummy_device`, `walkthrough/test_08_c10d`, `test_privateuse1_contract` probes); they fail identically on `dev` on this machine.
* `lintrunner -m origin/dev` → no lint issues.

Paths exercised: eager compile miss, C++ warm hit, and the Python path a view input takes. Not run: `torch.compile(backend="rbln")` graph mode (this path does not exist there), ATOM.

---

## Checklist

* [x] This PR is linked to a corresponding issue — see [Contributing to torch-rbln](https://github.com/RBLN-SW/torch-rbln/blob/main/docs/CONTRIBUTING.md)
* [ ] Relevant documentation has been updated (if applicable) — none affected

---

## Notes

**Tests added** (`test/rbln/test_internal_op_utils.py::TestOutTensors`): a host-latest `out` ends device-resident on the compile miss, the warm hit, and three alternating view-path calls; the warm hit binds alternating buffers and keeps its entry; the compile miss leaves exactly one output's worth of device memory alive after `out` is dropped (the first run's buffer is not retained); the Python path keeps one runtime per entry, one entry per scalar value, and none for grad mode; `add(x, x)` followed by `add(x, y)` of the same shapes gets its own graph instead of reusing the one-input one. Device residency is asserted through the runtime's hidden-copy counter (a clone of a host-latest source takes the slow path), not inferred from values.

**Regression and performance impact.** Not measured. Path comparison: on the compile miss the program runs twice instead of once plus a device copy. On a Python-path hit (view input, shim bail-out) the call skips Dynamo's guard evaluation, the runtime's output allocation and the copy into `out`; a host-latest `out` no longer leaves a host-to-device push for its next consumer. Warm hits through the C++ shim are unchanged.

**For reviewers.**
* `compile_and_run_view_aware` reads `runtime._input_binding.validated` to learn the graph's input order. It is a rebel-private attribute, the same tier as the `_runtime_holder` option this code already relies on; a public accessor on the runtime would be the right follow-up.
* If Dynamo does recompile inside a cached callable (a guard the key does not cover), `CompiledOp.runtime()` returns `None` and the call falls back to `compiled()` plus a device copy, with a one-time warning per entry. The key is meant to make this unreachable; the warning is how we would learn otherwise.
* A non-contiguous `out` (`out=base.t()`) still takes the runtime's own buffer plus a strided copy, as before.
* Binding a host-latest or `torch.zeros` `out` no longer pushes its stale contents: the pinned build carries rebellions-sw/rebel_compiler#13453, and `test_binding_out_moves_no_bytes_to_the_device` holds it to that.


torch_rbln/_internal/ops_utils.py
```diff
@@ -1856,29 +1867,88 @@ def _last_dim_unaligned(t):
...
+        # First call for this key: the backend builds the runtime while the
+        # callable runs, into a buffer of its own that is dropped here. Binding
+        # ``out`` costs a second run, once, under the compile that just happened.
+        external_result = entry.compiled(*view_args, **view_kwargs)
+        runtime = entry.runtime()
+        if runtime is not None and entry.input_order is None:
+            entry.input_order = _graph_input_order(runtime, tensors)
+        if out_list is not None:
+            if runtime is not None and entry.input_order is not None:
+                runtime.run(*(tensors[i] for i in entry.input_order), out=out_list)
```
rebel-huijong, line 1907: Using out= in runtime.run directly instead of doing external_result.copy_(result_tensor) when external_result.data_ptr() != result_tensor.data_ptr(). great!

rebel-huijong (approved):

PR #239, opened 2026-09-06
feat(vmem): keep device-resident storages on the device through the in-place host paths

rebel-chanheo:
## Summary of Changes

The runtime tracks sync state per vmem entry. A host borrow of a *view* therefore syncs the **whole storage** to the host and leaves it host-latest, and the next device consumer uploads all of it again. Three eager paths in torch-rbln paid that whole-storage round trip for a few bytes of work on a device-resident storage:

| path | before | after |
|---|---|---|
| `fill_` on a view of a device-resident storage (`fill_scalar_rbln_`) | borrow the host view: D2H of every byte, fill, storage becomes host-latest, next graph re-uploads it | enqueue one host pattern buffer against the view's contiguous runs on an `H2VBatch` (`enqueue_strided` with a zero source stride; the batch splits the submit to the runtime's bulk caps); the runtime routes by the entry's own state, so a device-resident storage is written in place and stays device-latest, a host-latest one is written in its user view with no transfer |
| boxed CPU fallback reading a small prefix view of a device-resident storage (`borrow_rbln_as_cpu`) | borrow reads back the whole storage to hand the kernel a few rows | views smaller than half the storage take the region copy path (v2h of the view when the data is on the device, a host memcpy of the view when it is not); the storage stays where it was |
| `copy_` into a non-contiguous device view from a shape/dtype-mismatched CPU source (`tensor_copy_from_cpu_to_rbln`) | stage the whole dst span through the host (read-modify-write) | convert the source on the CPU, then write the runs in place through the existing strided h2v plan |

**No runtime or compiler change is needed, and none of the paths asks the runtime where the data is.** Every call is one the cpu→rbln `copy_` path already makes, and the pinned runtime already routes an h2v copy by the entry's sync state (`PlanFastHostToVirtualCopy`, in `rebel-compiler` since 0.10.x): device-resident entry → written on the device in place, marked device-latest; host-latest / empty entry → written in the user view, no transfer. torch-rbln only stops borrowing (and thereby syncing) the whole entry.

Also worth knowing:

* The strided h2v plan keeps its existing minimum run size. A lower crossover would pay for device-latest destinations, but picking it needs the residency, which this PR deliberately does not query. No new configuration is introduced.
* Storages the device holds in a converted dtype (fp16/fp32 as dlf16) go through the runtime's converting user-view path for `fill_`, the same whole-entry cost as before; a device-side fill for them needs the runtime to accept a pattern in the physical dtype.

What this is not: it does not change the compiled path, dispatch, or the cost of any host-latest storage beyond a memcpy of a small view in the fallback case. Layer 3 keeps its entry-granular sync model; torch-rbln stops asking for the whole entry when it needs a region.

**Effect, measured on one RBLN-CR03 against the published rebel-compiler wheel (relative; the harness is `fill_micro.py`, attached in the PR discussion):**

* A CPU-fallback op reading a small prefix of a device-latest tensor: roughly one tenth of the previous cost, and the tensor stays on the device.
* An in-place `fill_` of a region of a device-latest buffer: about a fifth faster per call, and the storage is no longer pulled to the host and re-uploaded by the next graph.
* A region `fill_` of a host-latest tensor: several times faster than the borrow path.
* The size heuristic's worst case (a small view of a host-latest storage read by a fallback): a few percent slower, the memcpy of that view.
* Tiny fills and device-latest small fills: unchanged within noise.

End-to-end on the EAGLE3 spec-decode serve that motivated this (fsw-inference#548), step time is neutral at the median: these round trips are a small fraction of a step, and the remaining levers are in the model code (fsw-inference#1048, #1042). This PR is the structural fix for the eager side, not a step-time win by itself.

---

## Related Issues / Tickets

* Resolves rebellions-sw/torch-rbln-internal#40
* Related to rebellions-sw/fsw-inference#548 (EAGLE3 spec-decode device-tensor overhead: where the whole-storage round trips were found)

---

## Type of Change

---

## Affected Modules

- [x] Ops/Kernels (`fill_`, `copy_`, boxed CPU fallback)
- [x] Memory (`c10/rbln`)

---

## How to Test

Base for every number above: torch-rbln at this branch head; rebel-compiler `0.11.3.dev237+g687d3f76` (the newest published; the pinned dev185 has the same h2v routing); one RBLN-CR03. No runtime dependency beyond the pin.

1. **Against the published runtime (what CI sees).** Rebuild, then:
   ```bash
   RBLN_DEVICES=7 uv run --no-sync pytest test/rbln/test_fill_region_device.py test/rbln/test_fallback_region_io.py -v
   # -> 17 passed
   RBLN_DEVICES=7 uv run --no-sync pytest test/rbln/test_tensor_copy.py test/rbln/test_view_copy.py test/rbln/test_cpu_fast_paths.py test/rbln/test_strided_host_copy.py test/rbln/test_non_zero_storage_offset.py test/rbln/test_native_fill_scalar.py
   # -> 832 passed, 3 skipped (same as origin/dev)
   ```
   Both files carry `@pytest.mark.test_set_ci`, so they run on this PR. Against an `origin/dev` build (same wheel, same device) the six tests that cover the change fail — `test_tail_slice_single_run`, `test_strided_column_region`, `test_full_contiguous_device_latest_tensor`, `test_dtypes_native_physical`, `test_boxed_fallback_reads_only_the_view`, `test_strided_copy_from_mismatched_cpu_src_writes_only_runs` (e.g. `read back more than the 4-row view`) — and the eight edge-case guards pass on both, as they should.
2. **What the tests assert.** Each test builds a device-latest storage (`torch.compile(..., backend="rbln")` identity output), runs the in-place op on a view, and checks with `_C._rt_prof_host_sync()` deltas that (a) the result equals the CPU reference, (b) no D2H happened during the op and the H2D moved at most the region's bytes, and (c) reading the tensor back afterwards moves every byte from the device, i.e. the storage is still device-latest. Host-latest and fresh (never materialised) storages are covered too: nothing may move, and reading them back is a host copy.
3. **Edge cases covered by the tests:** fill on a converted-dtype storage (fp16/fp32: correctness only, whole-entry path as before); fill on a host-latest storage (nothing moves, stays host-latest); fill on a fresh `torch.empty` storage; a region larger than one bulk call may carry (split by the batch); an overlapping `unfold` view (declined: bulk destinations must be disjoint); a run count above 4096 (declines, borrow path); the fallback heuristic's worst case (a small view of a host-latest storage: no transfer, host memcpy only) and its boundary (a full view keeps the borrow); cpu src with a different dtype and shape into `x[::4]`; runs below the 16 KiB threshold (staged path, correctness only).
4. **Reproducer without the suite** (any host with an RBLN device):
   ```python
   import torch, torch_rbln
   from torch_rbln import _C
   f = torch.compile(lambda t: torch.maximum(t, t), backend="rbln", dynamic=False)
   x = f(torch.randn(4, 512, 768).to(torch.bfloat16).to("rbln"))     # device-latest
   torch.rbln.synchronize()
   (d0, b0), _ = _C._rt_prof_host_sync()
   x[:, 300:, :].fill_(0.0); torch.rbln.synchronize()
   (d1, b1), _ = _C._rt_prof_host_sync()
   x.cpu(); torch.rbln.synchronize()
   (d2, b2), _ = _C._rt_prof_host_sync()
   print("D2H bytes during fill_:", b1 - b0, "| D2H bytes reading back:", b2 - b1)
   # before this PR: the whole storage, then 0 (it was already on the host)
   # after:          0, then the whole storage (the data stayed on the device)
   ```

---

## Checklist

* [x] PR title follows Conventional Commits
* [x] Linked to a corresponding issue (fsw-inference#548)
* [x] Linting passes — `uv run --no-sync lintrunner -m origin/dev -a` in a fresh worktree venv (ruff, clang-format, clang-tidy against a real `compile_commands.json`)
* [ ] All CI tests pass
* [x] Test method and expected results described above
* [x] No documentation change needed: no new configuration or public API; the behaviour is described in the code comments at the three sites

---

## Notes

* **Why no residency query.** An earlier draft asked the runtime for the entry's sync state through a new C API. That API is not in any published rebel-compiler, and it turned out to be unnecessary for the main path: the runtime's h2v copy already routes by that state. So `fill_` never asks, the fallback uses a size heuristic (below half the storage → copy path; worst case a host memcpy of a small view), and the strided threshold keeps its existing constant. The exact residency-aware variants of the last two are a possible follow-up once such a query exists; the gain on the motivating workload came from `fill_`.
* **Review focus:** (1) `fill_region_device` in `RBLNTensorFactories.cpp`: the overlap guard (same rule as `make_strided_host_plan`), the run enumeration (`contig_suffix_start`) and the caps (≤4096 runs, ≤16 MB pattern); anything else falls to the previous path. The submit goes through `H2VBatch`, which enforces the runtime's 16-entry / 8 MiB bulk caps. (2) `borrow_rbln_as_cpu`: the new early-return sits before the storage_offset guard and routes to the existing region copy path. (3) `RBLNCopy.cpp`: `strided_plan_possible_for_contiguous_src` decides from the dst alone whether preparing a contiguous source can pay off, so the conversion happens at most once, and the staged path reuses it when the plan still declines. The profiler attribution follows the work actually done: the src conversion is recorded as `copy_h2d_staging`, and `copy_h2d_noncontig_dst` is recorded only when the dst really is pulled to the host (`test_strided_copy_profiler_attribution`).
* The relative figures above come from `fill_micro.py` (best of several batches per case, both builds against the same wheel on the same device); the raw numbers stay internal per the repository's publication rule.
* Debug-level log strings (`RBLN_LOG_DEBUG`) are compiled out in Release, so grepping the built `.so` for them says nothing; `nm -C libtorch_rbln.so | grep strided_host_copy` / the new test results are the evidence the rebuild took.

### DIFF
--- a/aten/src/ATen/native/rbln/RBLNCPUFallback.cpp
+++ b/aten/src/ATen/native/rbln/RBLNCPUFallback.cpp
@@ -204,6 +204,15 @@ at::Tensor borrow_rbln_as_cpu(const at::Tensor& t, uint64_t& borrow_id_out) {
     // host region and corrupt reads. Caller falls back to the copy path.
     return {};
   }
+  // A borrow syncs the whole vmem entry, so borrowing a small view of a device-latest
+  // storage reads back every byte of it and leaves the storage host-latest. The copy path
+  // moves only the view: a D2H when the data is on the device, a host memcpy when it is
+  // not. Without asking where the data is, views under half the storage take the copy
+  // path: the worst case costs a memcpy of a small view, the best case saves the
+  // whole-storage round trip. Larger views keep the zero-copy borrow.
+  if (static_cast<size_t>(t.numel()) * t.element_size() * 2 < t.storage().nbytes()) {
+    return {};
+  }
   if (t.storage_offset() != 0) {
     // Contiguous offset view (e.g. `x[1:]`): interior borrows ARE offset-correct.
     // The runtime's BorrowVirtualAtHost bounds-checks [vaddr, vaddr+size) against
--- a/aten/src/ATen/native/rbln/RBLNCopy.cpp
+++ b/aten/src/ATen/native/rbln/RBLNCopy.cpp
@@ -87,6 +87,29 @@ struct IndirectD2DGuard {
 // the errors are asymmetric: staging a slab descriptors would have won costs ~1.3x,
 // descriptors on element-sized slabs 130x.
 constexpr size_t kMinStridedHostCopyBytes = 16 * 1024;
+// Below the crossover the plan declines and copy_ stages through the host. For a
+// device-latest dst that staging is a read-back and rewrite of the whole span, so a lower
+// crossover would pay off there; choosing it needs the residency, which this path does not
+// query, so the constant stays at the host-latest crossover.
+
+// Whether make_strided_host_plan would accept `dst` paired with a contiguous source of the
+// same shape and dtype: the inner block is then dst's own contiguous suffix and the overlap
+// rule is dst's, so the answer does not need the source to exist yet.
+bool strided_plan_possible_for_contiguous_src(const at::Tensor& dst) {
+  if (dst.numel() == 0 || dst.is_contiguous()) {
+    return false;
+  }
+  if (at::has_internal_overlap(dst) != at::MemOverlap::No && view_may_self_overlap(dst.sizes(), dst.strides())) {
+    return false;
+  }
+  const auto sizes = dst.sizes();
+  const int64_t inner_start = contig_suffix_start(sizes, dst.strides());
+  int64_t inner_block_elems = 1;
+  for (int64_t i = inner_start; i < dst.dim(); ++i) {
+    inner_block_elems *= sizes[i];
+  }
+  return static_cast<size_t>(inner_block_elems) * static_cast<size_t>(dst.element_size()) >= kMinStridedHostCopyBytes;
+}
 
 // Byte-level description of one strided pair. `inner_block_bytes` is the largest
 // run contiguous in both tensors; the outer vectors describe the iteration above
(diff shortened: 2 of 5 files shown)


rebel-chanheo:
With a release scheduled for this weekend, I will merge this first to reduce overhead concentrated at the end of the month. Please provide feedback if you have any suggestions later.

PR #242, opened 2026-09-07
feat(compile): expose torch.rbln.capture_programs()

rebel-jonghewk:
## Summary of Changes

torch.compile keeps the callable a backend returns inside dynamo's code cache, so a caller never receives a handle to what was compiled. vllm-rbln currently reaches the built runtimes through rebel-compiler's private `_runtime_holder` option.

This PR exposes rebel-compiler's new `capture_programs()` scope on the `torch.rbln` namespace:

```python
with torch.rbln.capture_programs() as programs:
    compiled = torch.compile(model, backend="rbln")
    compiled(*warmup_inputs)          # dynamo builds the graphs here

for program in programs:              # list[torch.rbln.CompiledProgram]
    program.name                      # dynamo compile id, e.g. "0/0"
    program.device                    # rbln device the runtime is bound to
    program.runtime                   # execution handle behind the callable
```

Layer: torch-rbln Python only. `torch_rbln/programs.py` re-exports `capture_programs` / `CompiledProgram` from `rebel.core.program_capture`, and `torch_rbln/device/__init__.py` puts them on `torch.rbln`. The scope itself (ordering, nesting, thread-locality) lives in rebel-compiler next to the backend that builds the programs, and is tested there; this PR adds the surface test and a graph-mode test on the rbln backend.

Depends on rebel-compiler rebellions-sw/rebel_compiler#13494 (`rebel.core.program_capture`). The pin in `constraints-build-dev.txt` moves to `0.11.3.dev309+gb72212c9.prod`, the first dev build that carries it (`#13494` is `dev308`).

Follow-ups (not in this PR): `CompiledProgram.compiled_model` / `.executor` accessors, `CompiledProgram.input_specs` (per-input device alloc contract), and the vllm-rbln migration off `_runtime_holder`.

---

## Related Issues / Tickets

* Related to rebellions-sw/rebel_compiler#13494

---

## Type of Change

- [ ] `fix` — Bug fix

---

## Affected Modules

- [ ] Backend
- [ ] Device
- [ ] Ops/Kernels
- [x] `torch.compile`
- [ ] Memory
- [ ] Build/Tools
- [ ] Docs

---

1. Run `uv run --no-sync pytest test/rbln/test_capture_programs.py -x`
2. Verify output: 4 passed (2 surface tests, 2 graph-mode tests on `rbln`). Verified on RBLN-CR13 `rbln:0` against a rebel-compiler `dev` build that carries `rebel.core.program_capture`.
3. Regression impact: `torch.compile(backend="rbln")` builds run one extra list append per backend invocation when a scope is open, nothing when none is; eager dispatch is untouched. Not measured.


rebel-chanheo (approved):
LGTM

rebel-chanheo:
One test fails because it is currently a known issue. There should be no problem with this PR merge.

`test_dummy_compiled_prefill_decode_runs_on_real_npu`

torch_rbln/programs.py
```diff
@@ -0,0 +1,26 @@
...
+        program.device  # rbln device the runtime is bound to
+        program.runtime  # the execution handle behind the callable
+
+Every open scope receives each program, so nested scopes see the same programs; an
+inner scope simply sees fewer. The scope is thread-local. The implementation lives in
+rebel-compiler next to the backend that builds the programs; this module is the
+``torch.rbln`` surface for it.
+"""
+
+from rebel.core.program_capture import capture_programs, CompiledProgram
```
rebel-minkyu: I guess we used to import the rebel package lazily on purpose, but with this change it looks like `import torch` will import rebel as well. Would it be okay?

rebel-chanheo: Agree. On main `import torch` loads no `rebel.*` module — every existing rebel import in torch_rbln is function-local. This top-level import, star-imported from `device/__init__.py`, puts the whole `rebel` package (dynamo backend + custom op registration) inside torch's autoload hook.

Suggest a module `__getattr__` (PEP 562) in `programs.py` that imports `rebel.core.program_capture` on first access and returns the rebel objects as-is, with `device/__init__.py` delegating to it instead of the star-import. A test asserting `rebel` stays out of `sys.modules` after `import torch` would pin the rule.

torch_rbln/programs.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+__all__ = ["capture_programs", "CompiledProgram"]
```
rebel-chanheo: Resolve the two names on first access instead of at import time, so `import torch` keeps `rebel` off its path (see the thread above).

```suggestion
from typing import Any


__all__ = ["capture_programs", "CompiledProgram"]  # noqa: F822  # resolved by __getattr__


def __getattr__(name: str) -> Any:
    if name in __all__:
        from rebel.core import program_capture

        return getattr(program_capture, name)
    raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
```

torch_rbln/device/__init__.py
```diff
@@ -3,3 +3,4 @@
 from torch_rbln.device.streams import *  # noqa: F403
 from torch_rbln.memory import *  # noqa: F403
 from torch_rbln.profiler import *  # noqa: F403
+from torch_rbln.programs import *  # noqa: F403
```
rebel-chanheo: The star-import resolves every name in `__all__` at import time and would defeat the lazy `__getattr__`, so delegate instead.

```suggestion
from torch_rbln import programs as _programs


def __getattr__(name: str):
    # torch.rbln.capture_programs / CompiledProgram: resolved lazily, see torch_rbln/programs.py.
    return getattr(_programs, name)
```

rebel-chanheo (commented):
Suggestions for the lazy resolution discussed above.

PR #245, opened 2026-09-08
fix(test): track vllm-rbln ci/torch-rbln-model-tests and keep torch-rbln under test

rebel-chanheo:
## Summary of Changes

The model suite installs vllm-rbln from a branch frozen in June. That branch hands the attention op a per-partition `seq_lens` tensor, which the pinned rebel-compiler rejects on RBLN-CR13 (`Batch decode requires seq shape [1, 1]`), so every graph-mode and dummy-compile case fails on that lineup. Nothing in torch-rbln is at fault; the pin is stale.

Moving the pin to current vllm-rbln changes the dependency set, and the first CI run on this PR showed what that costs. Each of those is addressed here as well.

* `tools/test/install-test-deps.sh`
  * default `VLLM_RBLN_REF` is `origin/ci/torch-rbln-model-tests` (vllm-rbln `dev` plus the fixes the suite needs).
  * vllm-rbln and, through its dependencies, optimum-rbln go into `.venv-inference`, which sees the test venv's site-packages through a `.pth` file (torch, torch-rbln, rebel-compiler, pytest shared). The `.pth` calls `site.addsitedir`, so an editable torch-rbln or an external rebel-compiler reaches it too, and the install is verified afterwards: torch, torch-rbln and rebel must resolve to the test venv's files, transformers must not. Both pin transformers 5; `test_transformers.py` loads EXAONE-3.5's hub code, which no transformers 5 release runs, so the test venv keeps transformers 4. Until now the suite avoided the clash by letting optimum-rbln 0.10 downgrade torch to 2.10 under a torch-rbln built for 2.11.
  * the editable install is `--no-deps` and `torch-rbln`/`torch` are excluded from the dependency list: vllm-rbln bounds `torch-rbln` to a release line the nightly under test does not satisfy, so a resolving install would replace it.
* `test/run_tests.py`: `test_transformers.py` runs with the test interpreter, `test_optimum_llm.py` and `test_vllm_llm.py` with the inference venv, read from the same `INFERENCE_VENV` the installer honours.
* `test/models/test_vllm_llm.py`: models are passed as the cached snapshot directory when the hub is offline (huggingface_hub 1.x validates a cached snapshot against the repo tree and the shared caches hold only the files inference needs); the compile-only checks accept the mega-cache bundle vllm-rbln writes now, and compare artifacts by content, since a recompile lands in the bundle at the path it already used.
* `constraints-build-dev.txt`: rebel-compiler `dev238`. Compile-only writes its artifacts through `rebel.core.mega_cache.write_bundle`, which exists from `dev234`; `#193` targets the same build, `#244` a newer one.

Layer: test tooling and the compiler pin. The seq-shape assertion is settled on the vllm-rbln side.

---

## Related Issues / Tickets

* Resolves https://github.com/rebellions-sw/torch-rbln-internal/issues/59
* Failing runs: rebellions-sw/fsw-integration 34174280892 (CR13, stale pin) and 34199534687 (first run of this PR).
* Related to rbln-sw/vllm-rbln `ci/torch-rbln-model-tests` (`db9f0025`), and to #193 / #244 for the compiler pin.

---

## Type of Change

- [x] `other` — test tooling / CI pin

---

## Affected Modules

- [x] Build/Tools

---

## How to Test

1. `bash tools/test/install-test-deps.sh` in a venv holding the torch-rbln nightly: `torch` and `torch-rbln` unchanged, transformers 4 in the venv, `.venv-inference` with vllm-rbln and optimum-rbln importable.
2. `python test/run_tests.py --suite=models`: `test_transformers.py` (EXAONE included), `test_optimum_llm.py` and `test_vllm_llm.py` pass; TP>1 cases skip on the REBEL lineup.

Done on an RBLN-CR03 host. CR13 and CA25 were not available locally; the CI run on this PR covers them. pypi.rbln.ai needs credentials this host does not have, so the local run used a mirror; CI reaches it directly.

---

## Checklist

* [x] PR title follows Conventional Commits
* [x] Linked to a corresponding issue
* [x] Linting passes (shellcheck with the repo's `enable=all`, flake8 rules on the Python files)
* [ ] All CI tests pass
* [x] Test method and expected result described
* [x] Relevant documentation updated (`docs/TEST_GUIDE.md`, script header)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,6 +30,7 @@ uv.lock
 # Virtual environments
 venv/
 .venv/
+.venv-inference/
 
 # Compiled objects & libraries
 *.o
--- a/constraints-build-dev.txt
+++ b/constraints-build-dev.txt
@@ -1,2 +1,2 @@
 # Build/dev dependency constraints (single source of truth).
-rebel-compiler==0.11.3.dev238+g510669fa.prod
+rebel-compiler==0.11.3.dev266+g140f8fa1.prod
--- a/constraints-build-dev.txt
+++ b/constraints-build-dev.txt
@@ -1,2 +1,2 @@
 # Build/dev dependency constraints (single source of truth).
-rebel-compiler==0.11.3.dev218+g8e76c496.prod
+rebel-compiler==0.11.3.dev238+g510669fa.prod
--- a/docs/TEST_GUIDE.md
+++ b/docs/TEST_GUIDE.md
@@ -199,7 +199,9 @@ python -m pytest test/rbln/test_graph_eager_mode.py -s -v -x
 #### Installing Test Dependencies
 
 ```bash
-# Install everything needed to run the full suite
+# Install everything needed to run the full suite. vllm-rbln and optimum-rbln
+# land in .venv-inference (they pin transformers 5; test_transformers.py needs
+# 4), and run_tests.py runs test_optimum_llm.py / test_vllm_llm.py with it.
 ./tools/test/install-test-deps.sh
 
 # Preview what would be installed (dry-run)
--- a/test/models/test_vllm_llm.py
+++ b/test/models/test_vllm_llm.py
@@ -12,7 +12,8 @@ hard-coded expected strings.
 
 Environment requirements
 ------------------------
-* ``vllm-rbln`` installed on ``origin/device_tensor_rebased`` (or descendant).
+* ``vllm-rbln`` installed on ``origin/ci/torch-rbln-model-tests`` (what
+  ``tools/test/install-test-deps.sh`` checks out).
 * ``vllm_rbln`` / ``vllm`` importable.
 
 Matrix
@@ -365,7 +366,11 @@ LLM(
     enforce_eager=False,
     gpu_memory_utilization=0.1,
 )
-arts = glob.glob(os.path.join(os.environ["VLLM_CACHE_ROOT"], "rbln", "**", "*.rbln"), recursive=True)
+root = os.path.join(os.environ["VLLM_CACHE_ROOT"], "rbln")
+# vllm-rbln writes either per-graph .rbln files or a single mega-cache bundle.
+arts = glob.glob(os.path.join(root, "**", "*.rbln"), recursive=True) + glob.glob(
+    os.path.join(root, "**", "mega_cache.bin"), recursive=True
+)
 assert arts, "compile-only produced no .rbln artifacts"
 print(f"OK artifacts={len(arts)}")
 """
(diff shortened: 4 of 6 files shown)


rebel-chanheo:
Review follow-up, all four in `b7f57b2`.

**1. The child could get its own torch (P1).** Confirmed: with a `.pth`-linked parent, `uv pip install --python <child> vllm==0.24.0+cpu` plans `+ torch==2.11.0+cpu` into the child. pip does not, because it reads the target interpreter's `sys.path`. The stack now installs with the inference venv's own pip whatever `UV` is set to, and `verify_shared_packages` checks the result: torch, torch-rbln and rebel must resolve to the same files as the test venv, transformers must not. It aborts the install otherwise. Exercised with `UV=1` from the repo root: exit 0, no torch in the child, `Shared with the inference venv: torch 2.11.0+cpu, torch_rbln 0.4.1.dev11+g27828b4, rebel 0.11.3.dev238+g510669fa`. The three outcomes (parent fallback, healthy, duplicate torch) were also driven through synthetic venvs.

Two things came out of writing that check. The probe runs with `-I` from a neutral directory and compares real paths, because a torch-rbln checkout has a `torch` symlink at its root and the installer normally runs from there; a naive path comparison rejects a healthy environment. And the embedded Python is compiled in review, since `bash -n` does not look inside a heredoc.

**2. Editable installs did not reach the child (P2).** Confirmed against a checkout whose torch-rbln is installed editable: the bare-path `.pth` gave `RuntimeError: Failed to load the backend extension: autoload`, since the parent's own `.pth` files never ran. The `.pth` now calls `site.addsitedir`, which processes them; the same environment then imports `<checkout>/torch_rbln/__init__.py`. It appends, so the inference venv's transformers still wins, and the verification asserts that.

**3. `INFERENCE_VENV` ignored by the runner (P3).** Both sides read it now; `.venv-inference` is only the default.

**4. The mega-cache test could not see a recompile (P4).** Correct, and my earlier commit is what weakened it. Artifacts are compared by `(size, mtime_ns, sha256)` per path instead of by name. Replaying `os.replace` on the bundle: the name-set check passes (misses it), the content check catches it, and an unchanged bundle raises nothing.

Verified on RBLN-CR03 with the installer run end to end from the repo root under `UV=1`, then `test_transformers.py -k exaone` (2 passed, test venv, transformers 4) and `test_vllm_llm.py` graph + dummy-round-trip (2 passed, inference venv). CR13 and CA25 are on this PR's CI.

rebel-chanheo:
The CR13 lane cannot report on this: since about 15:30Z every job on the `fast-3` runners sticks in `Initialize containers` and is cancelled after two hours (`pod failed to come online ... phase status Pending: {"code":404}`, `Cannot read properties of null (reading 'jobPod')`). It affects both `fast-3-c32m256r2` and `fast-3-c128m1024r8`, and the same pool ran this suite normally earlier today, so it is the runners rather than this branch.

CA25 is green here, and I ran what CI runs on an RBLN-CR03 host against `b7f57b2` to cover the rest. No failures in any suite.

| suite | result |
| --- | --- |
| core | 4 groups passed |
| ops | 1 group passed |
| distributed | 1 group passed (2 NPUs, so the cases needing 4 or more skip) |
| models | 3 groups passed: `test_transformers.py` 11 passed 1 skipped, `test_optimum_llm.py` 18 passed 10 skipped, `test_vllm_llm.py` 10 passed 2 skipped |

`models` went through `run_tests.py --suite=models`, so it exercised the installer, the two interpreters and the shared-package verification rather than a hand-built environment. Collection matches CI at 52 tests; only the pass/skip split differs by lineup (CR13 29/10/13 before this PR, CA25 41/7/4, CR03 39/0/13).

Still only CI can answer for CR13: the assertion this PR is about fires in the compiler's in-memory attention route, which CR03 does not take.

PR #247, opened 2026-09-08
fix(c10d): keep a byte-identical device view for byte-copy collectives

rebel-chanheo:
## Summary of Changes

`ensureRcclRawMemory` re-binds every byte-copy operand flat through `bind_device_memory` before each collective. The runtime treats a flat allocation and a with-transform view as different layouts, so for an operand that is already device-resident in the layout the compiler chose, the re-binding is not a no-op: the buffer is re-materialized through the host (device-to-host, reallocate, host-to-device) before RCCL touches it.

A compiled graph's static output is such an operand, and it is what a vLLM pipeline-parallel stage sends to the next stage. The graph re-applies its own layout on the next run, so the host round trip repeats on every step: once inside the send, once inside the graph's output preparation. In a pipeline-parallel profile both show up as the dominant host cost of the handoff, several times what the same bytes cost from a persistent flat buffer.

This change keeps a device view that already holds the user's bytes 1:1 and hands it to RCCL as it is. The check reads the runtime's `MemoryInfo` (already fetched here): the view must have a physical dtype equal to the tensor's dtype, an element count that covers the whole storage, and the tensor's logical device must be a single NPU (a multi-NPU logical device may spread one view over several device areas, and the collective path reads one). Views without a device view, flat views, dtype-converting views, padded views and multi-NPU devices still take the existing `bind_device_memory` path. `bind_device_memory` keeps its flat contract; only its doc comment is updated to say what re-binding a differently laid-out region costs.

Limits: `MemoryInfo` does not expose the element order, so a view that permutes elements while keeping dtype and count would pass the check. No such view has been observed for byte-copy operands; the new test compares received bytes against the sender's host copy across several graph runs so that one would fail loudly rather than silently.

Layer: torch-rbln C++ extension (`ProcessGroupRBLN.cpp`), a doc comment in `c10/rbln/RBLNFunctions.h`, one new test. No runtime or Python change.

---

## Related Issues / Tickets

* Related to `rebellions-sw/torch-rbln-internal#61` — the analysis of the pipeline-parallel handoff cost that led to this change (internal). The public issue #246 was folded into it and closed.

---

## Type of Change

---

## Affected Modules

- [x] Memory

---

## How to Test

1. `pytest test/distributed/test_process_group.py -k compiled_output` — rank 0 compiles a small `Linear` with `use_static_output`, runs it three times and sends each output; rank 1 receives into one buffer and compares the bytes with the sender's host copy (sent over the CPU path of the same group). Passes with this change.
2. The existing `test_send_recv` matrix still passes: operands allocated with `torch.full`/`torch.zeros` have no with-transform view and keep the flat re-binding path.
3. In a two-rank micro-benchmark that sends a compiled graph's static output, `strace` shows one device job per send (the DMA) instead of three (device-to-host, host-to-device, DMA), and the send's host time comes down to what a send of a persistent flat buffer costs.
4. A MiniMax-M2.7 PP=4 vLLM prefill shows the same change on the per-handoff `rbln::send`/`rbln::recv` and on the runtime's output preparation, and the time-to-first-token improves accordingly. The first tensor's send now mostly reflects the sender's outstanding compute, which the DMA legitimately waits for.

---

## Checklist

* [x] Linting passes (`lintrunner -m origin/main`)
* [ ] All CI tests pass
* [x] Relevant documentation has been updated (doc comments)

---

## Notes

The root cause sits in the runtime's layout rule (a flat request is never considered compatible with a with-transform view, even a byte-identical one). Teaching the runtime that equivalence would let this check go away; until then the collective path is the right place for it, since it is the one caller that only needs the bytes.

### DIFF
--- a/c10/rbln/RBLNFunctions.h
+++ b/c10/rbln/RBLNFunctions.h
@@ -133,8 +133,10 @@ C10_RBLN_API bool is_eager_malloc();
  * shape a consumer that treats the region as bytes expects. Use set_device_layout_like()
  * instead when the region has to match another tensor's layout.
  *
- * Re-binding an already-bound region is allowed and is what the collective path does
- * before every operation.
+ * Re-binding an already-bound region is allowed. A region whose device view is already
+ * flat is left as it is; one that carries a different layout is re-materialized through the
+ * host, so callers that only need the bytes should check the view first (see
+ * ensureRcclRawMemory in the collective path).
  *
  * @param rbln_data Base pointer of an RBLN allocation (not an interior address).
  * @param nbytes Size of the allocation in bytes. Must be positive.
--- a/test/distributed/test_process_group.py
+++ b/test/distributed/test_process_group.py
@@ -256,6 +256,57 @@ def run_send_recv_test(rank: int, world_size: int, backend: str, src: int, dst: 
         dist.destroy_process_group()
 
 
+def run_send_recv_compiled_output_test(rank: int, world_size: int, backend: str) -> None:
+    """Send a compiled graph's static output and check the receiver gets the same bytes.
+
+    A compiled graph keeps its outputs in the layout the compiler chose, not as a flat
+    allocation, and re-applies that layout on every run. The byte-copy collective path must
+    hand such a view to RCCL as it is: re-binding it flat would re-materialize the buffer
+    through the host before every send. This test pins the correctness side of that
+    contract by comparing the received bytes with the sender's host copy across several
+    runs of the graph, so a layout the collective path must not treat as raw bytes shows up
+    as a mismatch here.
+    """
+    setup_environment(rank, world_size)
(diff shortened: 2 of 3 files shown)


rebel-chanheo:
Holding this as a draft: the fsw-integration ATOM distributed lane fails on this commit (`test_tp_pp.py` send/recv variants) while the same lane passes on the baseline runs just before it, so this is a regression from the check added here, not pre-existing flakiness.

Root cause of the limit: `MemoryInfo` exposes only dtypes and shapes, so the check here cannot see how many device areas back the view, nor the element order. A view that is byte-identical by dtype and element count can still be backed by several areas, and the collective path resolves a single one. That is information only the runtime has.

Next step is to move the equivalence decision into the runtime, where the area list and the replay are both visible, and drop this check. I will repurpose or close this PR once that lands.

PR #249, opened 2026-09-08
fix(warm-cache): scope empty_cache to a device, and stop rebuilding an entry whose hit failed

rebel-chanheo:
## Summary of Changes

Two WarmCache lifecycle bugs. The diff is a net deletion: the second one is fixed by removing the machinery that caused it, not by adding more.

**A failed hit rebuilt the same entry, forever.** When a hit failed at run time, the C++ side erased the entry and set a thread-local force-recompile flag; the Python wrapper consumed it, dropped its own compile-cache entry for the key, ran `torch.compile` again, and installed the result. A fresh compile of the same graph yields the same runtime, so the next hit failed the same way — one wasted recompile and one failed hit per dispatch of that key, for the life of the process.

The entry now stays where it is. The hit path already falls through to the pybind miss path on failure, whose `DynamoRuntime` handles the v-memory routing the C ABI path cannot, and `install` is a `try_emplace`, so the key the Python wrapper re-offers is a no-op. Nothing loops. A later call with the same profile gets to try the hit again, which is what this failure calls for: it belongs to the buffers one call brought — v-memory from an allocation path the C ABI cannot resolve — not to the entry, so a call that does not bring them should not be held back. `WarmCache::erase`, the force-recompile flag, its two pybind bindings and `compile_rbln_cached(force_recompile=)` are removed with it.

The cost of leaving a permanently unservable entry in place is one failed `rbln_rt_prepare_inputs` per dispatch, on a key that is already paying the pybind hop and a Dynamo guard check. Not measured.

**`empty_cache(device)` dropped every device's entries.** Serving several devices from one process, a flush on one of them put every other device's ops back on the Python wrapper path until each re-installed. `WarmCache::clear` now takes an optional device index and drops only the entries whose inputs live there, matching the allocator flush the same call already scoped that way. An entry with no device input at all would be dropped only by the no-device form; no shim op has one today, since each takes at least one input tensor, and the header says so.

Layer: torch-rbln C++ (`torch_rbln/csrc/`) and Python (`torch_rbln/`). Execution path: eager. Graph mode does not go through the warm cache.

---

## Related Issues / Tickets

* Resolves https://github.com/rebellions-sw/torch-rbln-internal/issues/62

---

## Type of Change

- [x] `fix` — Bug fix

---

## Affected Modules

- [x] Backend
- [x] Memory

---

## How to Test

1. `uv run --no-sync pytest test/internal/test_warm_cache_internals.py`

   `TestWarmCacheDeviceScope` warms one profile on each of `rbln:0` and `rbln:1`, calls `empty_cache(1)`, and asserts the cache size drops to exactly what `rbln:0` held, that `rbln:0` keeps hitting, and that `rbln:1` misses once and then hits again. It fails on `main`, where `rbln:0` also stops hitting. It skips below two devices.

   `TestWarmCacheForceRecompileFlag` is deleted along with the flag it covered. The rest of the suite is unchanged and covers the paths this touches: `TestWarmCacheHitPath` (a repeated profile keeps hitting) and `TestWarmCacheContractBreak` (a runtime the hit path cannot take a handle off stays harmless).

2. `uv run --no-sync pytest test/rbln/test_warm_cache_view_on_device.py test/rbln/test_dispatch_shim_precheck.py test/rbln/test_internal_op_utils.py`
3. `uv run --no-sync python test/run_tests.py --suite=core`

Two idle NPUs are needed for `TestWarmCacheDeviceScope`; it skips otherwise. fp16 throughout — fp32 falls back to CPU on this lineup and never reaches the warm cache.

---

## Checklist

* [x] Linting passes
* [ ] All CI tests pass
* [x] Relevant documentation has been updated — `empty_cache` docstring and the `warm_cache` module docstring

---

## Notes

**Not verified by the author.** The branch was not rebuilt and no test was run: the only editable install on this machine points at a working tree carrying unrelated uncommitted work. Everything under *How to Test* is the test's stated intent, not an observed run. `lintrunner -m origin/main` reports no issues except `CLANGTIDY`, which needs a compile database this tree does not have.

**No new path to review.** The failed-hit fix adds no branch — the `if (runtime_failed)` block already returned `false` into the miss path, and this drops the two calls it made before returning. Worth confirming that leaving an entry that can never be served is the right call: the alternative is retiring the key so the failed attempt is not repeated, which costs a tombstone in the cache, a new C++ entry point, and a test-only hook to fail a hit on demand. That seemed like more machinery than one failed C call per dispatch is worth, but it is a judgement call.


rebel-chanheo:
The `pytest - models` failure on this PR is the state of `main`, not this change: 7 failed / 41 passed / 4 skipped, the same set and the same counts as `main` before #245 (EXAONE under transformers 5, the two compile-only cases under a compiler without `mega_cache.write_bundle`, and two models whose offline snapshot check fails). This branch is cut from `main`, so it installs the harness #245 replaces.

The suites this change can affect are green here: core, ops, distributed and CTest.

PR #250, opened 2026-09-09
feat(c10d): reduce bfloat16 tensors natively through RCCL

rebel-chanheo:
## Summary of Changes

`ProcessGroupRBLN` reduced only float16 on the device. AllReduce and ReduceScatter sent every other dtype, bfloat16 included, through the host: copy to CPU, reduce with the Gloo backend, copy back. RCCL and the runtime already expose a bfloat16 reduce type, and the UMD test tool reduces bfloat16 natively, but torch-rbln never selected it.

bfloat16 now takes the same device path as float16:

* the storage is registered with bfloat16 as its physical dtype (float16 keeps its CustomFloat16 transform), and
* the collective runs with the RCCL bfloat16 type instead of the hard-coded CustomFloat16 type.

Keeping the dtype end to end also preserves the bfloat16 value range; a float16 intermediate would overflow at magnitudes bfloat16 represents. The dtype gate, the physical dtype and the RCCL type now come from three small helpers next to the element-size helper, so the reducible set is written down once.

Two side fixes in the same code:

* ReduceScatter registered the output with the reduce layout before checking the dtype, so a tensor headed for the host path was still given a CustomFloat16 physical view. The registration now follows the check.
* The AllReduce host path nested one loop over the inputs inside another and staged every input once per input. Each input is copied once now (separate commit).

Layer: C++ extension (`torch_rbln/csrc/distributed/c10d`). Compiled graphs are unaffected: their collectives are lowered in-graph by the compiler and never reach this process group.

---

## Related Issues / Tickets

* No linked issue. Found while auditing the dtype gate of the reduce collectives; the fallback was reachable from any eager `dist.all_reduce` / `dist.reduce_scatter` on a bfloat16 device tensor.

---

## Type of Change

---

## Affected Modules

---

1. On a host with two free NPUs, run the collective tests: `RBLN_DEVICES=<a>,<b> python -m pytest test/distributed/test_process_group.py -k "TestAllReduceRBLN or TestReduceScatterRBLN"`
2. Expect all tests to pass. New coverage: `test_allreduce_sum` runs for every dispatch dtype, `test_allreduce_range` checks that a bfloat16 sum beyond the float16 range stays finite and exact, and `test_allreduce_bfloat16` / `test_reduce_scatter_sum_bfloat16` sweep the single-call, unaligned-tail and chunked paths.
3. To see the path change directly, wrap an eager bfloat16 `dist.all_reduce` in `torch.profiler`: before this change the trace contains `gloo:all_reduce` and host copies under `rbln::all_reduce`; after it only `rbln::all_reduce` remains, and the call takes about as long as the float16 one instead of several times longer.

Run here on two REBEL NPUs: the AllReduce and ReduceScatter test classes pass in both sync and async c10d modes, including the existing float32 host-path test. Lint is clean against `main`.

---

## Notes

* float32 (and every other non-16-bit-float dtype) still takes the host path by design; the runtime has no device reduce for it.
* The host-path `reduce_scatter` is noticeably slower than the host-path `all_reduce` at the same byte count. That is unrelated to the dtype gate (moving the registration did not change it) and is left as is.


rebel-huijong (approved):
@rebel-minkyu 
We don't actually have much occasion to do reduce-scatter with torch-rbln, but this is a good diff for seeing the overall flow of using RCCL through torch-rbln, so it's worth a look. Right now, the one meaningful use case for RCCL via torch-rbln is the send/recv path in the vllm-rbln PP setup.

PR #256, opened 2026-09-11
build(deps): bump up to torch 2.13

rebel-minkyu:
## Summary of Changes

Moves the torch pin from `2.11.0+cpu` to `2.13.0+cpu`.

**C++20 for extension builds** — [#176659](https://github.com/pytorch/pytorch/pull/176659), [#178343](https://github.com/pytorch/pytorch/pull/178343), [#178662](https://github.com/pytorch/pytorch/pull/178662) (2.12)
- *Upstream:* PyTorch builds as C++20, `cpp_extension` compiles extensions with `-std=c++20`, and the exported `torch` CMake target carries `CXX_STANDARD 20`.
- *Us:* our `CMakeLists.txt` asked for C++17. Not forced by the imported target, but upstream no longer validates its headers at C++17.
- *Fix:* `CMAKE_CXX_STANDARD 17 → 20`.

**libkineto `DeviceInfo`/`ResourceInfo` lost their constructors** (2.13)
- *Upstream:* both structs are plain aggregates now.
- *Us:* `rbln_kineto_emitter.cc` called `emplace_back(deviceId, id, sortIndex, name)`. At C++17 this fails to compile; at C++20 parenthesized aggregate init compiles it, but `ResourceInfo` declares `id, sortIndex, deviceId, name`, so the old argument order lands in the wrong fields silently.
- *Fix:* explicit brace init in member order.

**`torch.testing._utils.freeze_rng_state` became accelerator-generic** (2.13)
- *Upstream:* it calls `get_rng_state()`/`set_rng_state()` on the current accelerator's device module instead of CUDA's.
- *Us:* `torch.rbln` had neither, so 51 OpInfo tests over seeded ops (`normal`, `uniform`, `randn`, `empty_strided`, SDPA dropout) failed with `AttributeError`. `torch/utils/backend_registration.py` lists both as required privateuse1 device-module API anyway.
- *Fix:* add `get_rng_state`/`set_rng_state`/`get_rng_state_all`/`set_rng_state_all` to `torch_rbln/device/device.py` (CUDA parity; the per-device default generator is created on first use, so nothing claims a device at import). `RBLNGeneratorImpl::get_state` returned uninitialized bytes and `set_state` was empty, so the state never actually round-tripped — implemented as seed+offset serialization. Two contract tests added.

**`arange` int64 step validation** (2.13)
- *Upstream:* `compute_arange_size` checks the truncated int64 step, so `arange(0, 64, 0.5, out=int64)` raises `ValueError("step must be nonzero")`; 2.11 CPU used to SIGFPE on it. A matching OpInfo error input was added.
- *Us:* `arange_start_out_rbln` sized with double and then checked `step.to<int64_t>() == 0`, filling 128 zeros — a silent wrong answer. It also returned an empty tensor instead of raising for `end < start`.
- *Fix:* size and validate through `at::native::compute_arange_size<scalar_t>` dispatched on the out dtype; both cases now match CPU.

**Dynamo `ID_MATCH` guard no longer reprs the value** — [#184796](https://github.com/pytorch/pytorch/pull/184796) (2.13)
- *Upstream:* `GuardBuilder.id_match_unchecked` reprs the type, not the value.
- *Us:* `monkey_patches.patch_dynamo_guard_repr()` is a backport gated on `< 2.13`, so it is dead code at this pin, and the four tests asserting the patch was applied fail.
- *Fix:* delete the backport, its module state and its teardown branch, and replace those tests with one contract test pinning the upstream behavior — `mark_static_address(param, guard=True)` forces an ID_MATCH guard and a `torch.Tensor.__repr__` spy shows no value repr. The trigger was confirmed to repr at 2.11 and not at 2.13. The patch list in `AGENTS.md` is updated.

**Vendored file re-sync**

- `tags.yaml`: upstream added `reduction` (2.9) and `inplace`/`out` with `out_variant` deprecated (2.13). Re-synced; the diff against upstream is now the header line and our `use_custom_kernel_rbln` block only.
- `native_functions.yaml`: header to v2.13.0, `reduction`/`pointwise` tags aligned with upstream for the ops we carry. The signatures of all 49 entries are unchanged across v2.8.0–v2.13.0, and the generated `register_ops.py` is byte-identical before and after the sync.
- `test/ops/test_ops.py`: 3-way merge with v2.13.0 — `skip`/`skipOps`/`xfail` moved to `common_device_type`, `skipOps` takes one set, `overloads()` → `op_overloads()`, named-tensor overloads removed.
- Version strings in `README.md`, `docs/WORKFLOWS.md`, `tools/build-with-external-rebel.sh`, `tools/sync-linter.sh`.

---

## Related Issues / Tickets

* Resolves #
* Related to #

---

## Type of Change

- [ ] `fix` — Bug fix

---

## Affected Modules

- [ ] Backend
- [ ] Device
- [ ] Ops/Kernels
- [ ] `torch.compile`
- [ ] Memory
- [ ] Build/Tools
- [ ] Docs

---

1. Run `...`
2. Verify output: `...`
3. Edge case tested: `...`

---

---

## Checklist

* [ ] PR title follows [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) format
* [ ] Linting passes — see [Linting](https://github.com/RBLN-SW/torch-rbln/blob/main/docs/LINTING.md)

---

## Notes

---


rebel-chanheo (approved):
LGTM

torch_rbln/device/device.py
```diff
@@ -296,6 +302,46 @@ def _maybe_exchange_device(device: int) -> int:
     return _exchange_device(device)
 
 
+_default_generators: Dict[int, torch.Generator] = {}
+_default_generators_lock = threading.Lock()
+
+
+def _default_generator(device: Union[int, str, torch.device, None]) -> torch.Generator:
+    index = _normalize_device(device).index
```
rebel-jonghewk, line 310: Is normalize_device same as ` _get_device_index`?

torch_rbln/device/device.py
```diff
@@ -296,6 +302,46 @@ def _maybe_exchange_device(device: int) -> int:
...
+def set_rng_state(new_state: torch.Tensor, device: Union[int, str, torch.device] = "rbln") -> None:
+    """Set the RNG state of ``device`` from a ByteTensor produced by :func:`get_rng_state`."""
+    _default_generator(device).set_state(new_state)
+
+
+def set_rng_state_all(new_states: Iterable[torch.Tensor]) -> None:
+    """Set the RNG state of every device from states ordered like :func:`get_rng_state_all`."""
+    for index, state in enumerate(new_states):
+        set_rng_state(state, index)
+
```
rebel-jonghewk, line 343: I don't know alot about torch-rbln's code arch but should rng be in device.py? or is it better to separate as random.py ?

torch_rbln/_internal/monkey_patches.py
```diff
@@ -146,56 +141,9 @@ def reset_wrapper(*args, **kwargs):
         _torch_dynamo_reset_patched = True
 
 
-def patch_dynamo_guard_repr() -> None:
```
rebel-jonghewk, line 149: Thanks for cleaning this up! 👍

torch_rbln/device/device.py
rebel-huijong: With the torch version bump, missing these functions now raises an AttributeError outright. So for the bump up, I've just put in placeholders to satisfy the interface for now as the torch-rbln package needs to ship soon so we can start preparing the SDKs that depend on it.

As for actually implementing this random functionality: we can't really support a random generator properly right now, so we'd need to keep a CPU generator and handle it as a fallback. There's an earlier draft PR where I was working on that, and my plan was to merge this one first and then pick that up next. Would you prefer I fold it into this PR instead?

torch_rbln/device/device.py
rebel-huijong: I'll check this right away.

rebel-huijong: Nice catch! Apart from `_normalize_device` doing a stricter range check, it isn't functionally that different from `_get_device_index`.

@rebel-chanheo 
Is there a history behind having two helpers that do essentially the same thing? If not, it would be better to consolidate this two.

rebel-chanheo: They overlap, but aren't equivalent. _normalize_device resolves None to the current device and validates booleans and device bounds, while _get_device_index preserves the -1 sentinel used by device-selection APIs. Replacing it directly would change behavior.

rebel-ywkim (approved):
lgtm!

torch_rbln/device/device.py
rebel-huijong: Ah, I see they're different. The -1 handling seems to be an important case and I missed it. Thanks for sorting it out.

torch_rbln/device/device.py
rebel-jonghewk: Got it. I think it will be okay for now.

PR #265, opened 2026-09-14
test(models): drop the 3-way interactions from the transformers grid

rebel-chanheo:
## Summary of Changes

Cut the transformers model grid from 54 release cases to 26, and from 12 CI cases to 6, by removing interactions and duplications that only cost time at the expensive end of the grid. Three commits, each with its own argument.

The release run spends almost all of its time in one corner: of its 54 cases, the 26 at seq 1024 with batch 2 or 4 account for roughly nine tenths of the wall time, and the other 28 together for less than a tenth. Every cut here is in that corner.

**1. The 3-way interactions.** `_small_model_cover_array` added a full dtype x attn corner at *two* shapes, (2, 1024) and (4, 1024), on both small models — a 3-way interaction on top of the pairwise rows. It now takes a single `full_matrix_at` shape, used for the CI representative point and carried by one model per family (Qwen2.5). The large models ran the full dtype x attn matrix at (2, 1024) *and* a separate largest-shape smoke, `test_large_model_tiling`; what a large model buys over its small sibling is geometry, not dtype/attn breadth, so they moved to the largest shape and that separate smoke was folded in.

**2. The duplication between the two small models.** Both compiled the whole 9-shape grid. The shape axis is about compiled artifacts, not about a particular model's geometry, so one small model can carry the costly shapes for both: Llama-1B skips (2, 1024) and (4, 1024), which Qwen2.5 compiles at both dtypes. It keeps (1, 1024), so a long sequence is still compiled on Llama geometry, and Llama-3B covers the family at the largest shape. The large models are likewise back to one case each, float16 and sdpa — exactly what their largest-shape smoke was — because both dtypes and both attns already run at that shape on the small models.

**3. The batch axis.** The small models ran batch 1, 2 and 4 at every sequence length. Batch 4 is where per-batch tiling and memory pressure show, and the large models already run it at the largest shape, so the small models keep batch 1 and 2 and batch 4 stops being paid for twice.

## Coverage

What still holds, re-checked against the collected test ids rather than asserted in prose:

- every sequence length is compiled at batch 1 and 2, and batch 4 at the largest;
- each small model covers every dtype x shape, attn x shape and dtype x attn pair over the shapes it runs (Llama-1B 10/10, 10/10, 4/4 over 5 shapes; Qwen2.5 12/12, 12/12, 4/4 over 6);
- the full dtype x attn matrix runs at the CI representative point.

What is deliberately given up, in the order a reviewer is likely to object:

- **batch 4 now runs in one combination only** — float16 with sdpa, at seq 1024, on the two large models. Batch 4 at seq 16 and 128, and batch 4 with bfloat16 or eager, are gone;
- **the guarantee that every shape is compiled is suite-wide, not per model** — Llama-1B no longer compiles (2, 1024);
- the 3-way dtype x attn x shape away from the CI point;
- bfloat16 and eager attention on a large model. Both run at seq 1024 on the small models, at batch 2.

## Affected Modules

`test/models/test_transformers.py` only. `TestCausalLMPerf` (manual, `test_set_perf`) is untouched. Independent of #261 / #264 / #263, which touch other files, so it can merge in any order.

## How to Test

- Collection: 26 release cases (was 54), 6 CI cases (was 12); the coverage numbers above come from parsing those ids.
- Applying the per-case costs measured in the last release run to the new grid: the transformers suite drops from roughly an hour and a half to a bit over twenty minutes, and the CI models job by more than half.
- Executed on a REBEL host from the shared model cache: both large-model cases and small-model cases from the rotated rows pass.

---

## Related Issues / Tickets

* Related to rebellions-sw/torch-rbln-internal#65


rebel-chanheo:
I plan to perform the merge first and verify how much time is reduced through today's release. I'm sorry to ask for your understanding regarding the merge without a review.

PR #267, opened 2026-09-15
test: take the three slowest tests off rebel-compiler's critical path

rebel-chanheo:
## Summary of Changes

rebel_compiler runs this suite on every one of its pull requests under a 10-minute budget. Reading its CI data (obedients, 29 timed-out runs), the three most expensive tests it runs are below — and in each case the cost was not the thing under test.

| Test | CI cost (median, timed-out runs) | Where the time went |
|---|---|---|
| `test_empty_cache_releases_current_device` | 57.9s | `buf.add_(1.0)` on a 16M-element tensor, which compiles a kernel. `empty_cache` itself is immediate. |
| `test_boolean_spellings` | 37.8s | Twelve interpreters, one per spelling, started one after another. |
| `test_device_map_malformed_rejected` | 34.6s | Nine interpreters, one per malformed string, started one after another. |

**The allocation, not the arithmetic.** An empty tensor's device allocation is lazy until something materializes it, and this test needs the physical memory so that freeing it leaves a *reserved* block for `empty_cache` to release. `torch.rbln.bind_device_memory` is the API for exactly that; reaching it through an arithmetic op was incidental, and expensive at that size.

**One interpreter per case, but not one after another.** Both of the other tests genuinely need a fresh process per case: the dummy-device flag is registered `FlagMutability::Sealed` and `is_dummy_device()` caches for the process, and the device mapping is planned once per process. What they did not need was to pay for `import torch` twelve and nine times in sequence. The cases are independent, so they now run at once, bounded to eight at a time so the children's memory stays sane. Threads are enough — every call blocks in `subprocess.run`.

Same cases, same assertions, same process isolation. No test was removed or weakened.

## Affected Modules

`test/rbln/test_accelerator_contract.py`, `test/rbln/test_dummy_device.py`, `test/rbln/test_device_mapping.py`.

## How to Test

Development host, `RBLN_DEVICES` pinned to two idle NPUs:

| | before | after |
|---|---|---|
| `test_accelerator_contract.py` (whole file) | 35.0s | 3.5s |
| `test_boolean_spellings` | 20.3s | 3.1s |
| `test_device_map_malformed_rejected` | 18.0s | 3.2s |
| the three files together | 147.5s | 92.1s |

Pass/skip/xfail counts are identical before and after (70 passed, 17 skipped, 1 xfailed, plus four `TestMemGetInfo` failures that predate this branch — they need a newer rebel-compiler than the one in the venv used for the measurement).

## Notes for the reviewer

The parallel run uses `concurrent.futures` directly at each call site rather than a shared helper in `test/utils.py`. That is deliberate: `run_test_with_env` builds a child script that re-imports the test module, so a new cross-module import in `test_device_mapping.py` becomes a way for every subprocess test in that file to fail at once. A first draft did exactly that and broke seven tests.

---

## Related Issues / Tickets

* Related to rebellions-sw/torch-rbln-internal#65


rebel-huijong (approved):
For test_device_map_malformed_rejected and test_boolean_spellings is the approach here is to run the slow ones concurrently? And for test_empty_cache_releases_current_device, remove unnecessary compilation, since we only need physical memory?

PR #268, opened 2026-09-15
fix(log): log the CPU fallbacks that left no trace at any level

rebel-chanheo:
## Summary of Changes

`AGENTS.md` makes grepping the log the way to tell a run that fell back from one that did not. Two groups of CPU fallbacks never wrote a line, at any `TORCH_RBLN_LOG_LEVEL`.

**The C++ dispatch shim's pre-check.** It decides the dtype / NaN-Inf / all-scalar fallbacks for the ops it covers and called `cpu_fallback_rbln` directly, skipping the `log_cpu_fallback` that the `fallback_rbln` handler makes for an unsupported op. Those are the common ops — `add`, `sub`, `mul`, `div`, `matmul`, the reductions — so an fp32 workload ran every one of them on the CPU with correct values and an empty log:

```python
a = torch.tensor([1, 2, 3], dtype=torch.float32, device="rbln")
b = torch.tensor([4, 5, 6], dtype=torch.float32, device="rbln")
c = a + b          # falls back; nothing logged, even at TORCH_RBLN_LOG_LEVEL=INFO
```

It now calls the same helper, at the same point, as the handler it shortcuts past.

**Three `is_cpu_fallback_cases` branches in the SDPA kernel** returned into a CPU helper without logging, while their siblings on the same path (`sdpa_overrideable`, `sdpa_backward`) log. They now log too.

The line stays at **INFO**, where the existing ones are: fp32 is a supported way to run on this hardware, and raising it to WARNING would put a line in every such user's console. What changes is that `TORCH_RBLN_LOG_LEVEL=INFO` shows all of the fallbacks instead of some.

`log_cpu_fallback` now reads the level before it builds the message. The shim calls it on every fallback of the ops it covers, so at the default level the concatenation ran once per `add` only to be dropped by the sink. The warning half was already gated.

Coverage after this change: the 31 generated shims in `register_ops.py` log through `cpu_fallback_path`, the shim pre-check and the unsupported-op handler log through `log_cpu_fallback`, the SDPA kernel logs on all six of its branches, and the whole-graph compile fallback logs at WARNING in `torch_compile_patch_helpers.py`.

Layer: torch-rbln C++ (`torch_rbln/csrc/`, `c10/rbln/`) and Python (`torch_rbln/`). Execution path: eager.

---

## Related Issues / Tickets

* Related to #249 — found while auditing the warm-cache fallback paths there.

*(No tracking issue yet; happy to file one if this should have its own.)*

---

## Type of Change

- [x] `fix` — Bug fix

---

## Affected Modules

- [x] Ops/Kernels

---

## How to Test

```bash
uv run --no-sync pytest test/rbln/test_cpu_fallback_logging.py
```

Four scenarios, each in its own subprocess: `TORCH_RBLN_LOG_LEVEL` is latched in a function-local static on first use, so it cannot be changed within a process.

* `test_shim_op_dtype_fallback_is_logged` runs the snippet above at `INFO` and asserts the line names `aten::add`.
* `test_fallback_log_stays_off_by_default` pins the level, asserting the line stays out of a `WARNING` run.
* `test_sdpa_inner_fallback_is_logged` puts a NaN in the inputs to trip the two per-step checks in the forward.
* `test_sdpa_backward_fallback_is_logged` puts a NaN in the incoming gradient to trip the one in the backward.

Verified against the `main` behaviour by reverting each half in place: with the shim's call removed and the three SDPA calls made no-ops, three of the four fail; with them restored, four pass.

---

## Checklist

* [x] Linting passes
* [x] All CI tests pass
* [ ] Relevant documentation has been updated (n/a)

---

## Notes

**Open question — no deduplication.** At `INFO`, a loop of fp32 `add` prints one line per call. Both paths reach the same helper, so a per-op-name once-only gate would be one place; counting occurrences is the reason not to. Left as-is, matching what the Python path has always done.

**Also found, not fixed here:** `TORCH_RBLN_DISABLE_FALLBACK=non_blocking_copy` is documented in `docs/CONFIGURATION.md` and validated by `RBLNFallbackConfig.cpp`, but nothing calls `is_fallback_disabled("non_blocking_copy")` — setting it does nothing, silently. Separate from this change.


rebel-minkyu (approved):
LGTM!

PR #274, opened 2026-09-16
test(out): drop the guard against rebel's out-tensor stash

rebel-chanheo:
## Summary of Changes

`TestOutTensors::test_out_is_filled_without_the_rebel_global_out_stash` opens with
`from rebel.core.torch_eager import eager_execution_helper` and patches `set_out_tensor`
to assert that torch-rbln never reaches for rebel's process-global out-tensor stash.

#255 removed the last torch-rbln use of that stash. rebel is now removing the module
itself — nothing sets the stash any more, so `DynamoRuntime.run(out=)` is the only way a
caller's buffer reaches the runtime. Once that lands, this test's import raises
`ImportError` and the name it patches no longer exists. The guard also stops being a test
result at that point: a call to the stash cannot be written at all, because the name it
would need is gone.

Everything else the test checked is covered where those paths are covered anyway:

| Assertion | Covered by |
| --- | --- |
| `result is out`, `out.data_ptr()` preserved | `test_add_using_contiguous_out` and its three contiguity / storage-offset variants |
| `out=` filled on the compile miss and on the warm hit | `test_host_latest_out_becomes_device_resident_on_every_path` (both paths explicitly), `test_warm_hit_binds_alternating_out_buffers` |
| the op took the device path, not a host fallback | the `assertFalse(_is_host_latest(out))` assertions in those two tests — a host-computed result leaves `out` host-latest, which is what the warm-cache-size assertion here stood in for — and `test_binding_out_moves_no_bytes_to_the_device`'s transfer counters |
| the stash is not used | this test only; enforced by the module's absence afterwards |

The imports the deleted method used (`unittest.mock.patch`, `_C._warmcache_size`) stay:
other tests in the same file use both.

This is ordering work, not a fix for a red lane. The pinned build
(`rebel-compiler==0.11.3.dev426+gd8aa19b2.prod`) still ships `rebel/core/torch_eager.py`,
so the test passes today; merging this first keeps the rebel-side removal from turning the
torch-rbln lane red when the pin next moves.

---

## Related Issues / Tickets

* Related to rebellions-sw/torch-rbln-internal#37
* Follows #255, which removed torch-rbln's uses of the stash
* Unblocks the rebel-side deletion of `rebel/python/rebel/core/torch_eager.py` (rebel_compiler branch prepared, not yet opened)

---

## Type of Change

- [x] `other` — test-only removal

---

## Affected Modules

Tests only (`test/rbln/test_internal_op_utils.py`); no module under `torch_rbln/` is touched.

---

1. `uv run --no-sync pytest test/rbln/test_internal_op_utils.py -k TestOutTensors`
   Expected: the six remaining `TestOutTensors` cases pass, and the deleted one is gone
   from the collection.
2. `uv run --no-sync lintrunner -m origin/main`

Run here: `ruff check` and `ruff format --check` on the file, plus a check that no
reference to `torch_eager` / `eager_execution_helper` / `set_out_tensor` is left anywhere
in the repository. The suite itself was not re-run — this deletes one test method and
touches nothing the others use; CI covers them.

---

## Checklist

* [x] This PR is linked to a corresponding issue — see [Contributing to torch-rbln](https://github.com/RBLN-SW/torch-rbln/blob/main/docs/CONTRIBUTING.md)
* [ ] Relevant documentation has been updated (if applicable) — none affected

---

## Notes

**For reviewers.** The question to settle is whether the stash guard is worth keeping in a
form that survives the module's removal. It is not, as far as I can tell: torch-rbln's
dependency range (`rebel-compiler>=0.11.2,<0.20.0`) does admit older wheels where the stash
still exists, but CI always runs a pinned recent build, so the guard would only ever run
against a rebel that has already lost the API.


rebel-chanheo:
the purpose of this PR is to remove the test for `eager_execution_helper()`  because we plan to remove helper code from rebel_compiler. this test code will cause fail from torch-rbln test in rebel-compiler's PR.

